PR #58, opened 2021-01-25
Fix ci config issues

otterdahl:
Fixes CI config issues for the ITC2 and EC2 simulators.

- [x] Uploads validation log
- [x] Uploads the internal rsmp log of the simulators


otterdahl:
Fixes CI config issues for the ITC2 and EC2 simulators.

- Uploads validation log
- Uploads the internal rsmp log of the simulators

PR #61, opened 2021-01-26
Added disruption and time sync tests

GunnarRolander:


spec/support/command_helpers.rb
```diff
@@ -296,6 +296,40 @@ def set_date
     send_command_and_confirm @task, command_list, "intention to set date"
   end
 
+  def verify_synchronization
```
emiltin, line 299: if this code is not reused between several tests, it should be placed directly in the rpsec test

spec/support/command_helpers.rb
```diff
@@ -296,6 +296,40 @@ def set_date
...
+      securityCode: SECRETS['security_codes'][1],
+      year: "2020",
+      month: "9",
+      day: "29",
+      hour: "17",
+      minute: "29",
+      second: "51"
+    }
+    send_command_and_confirm @task, command_list, "intention to set date"
+    request_status_and_confirm "current date and time",
```
emiltin, line 311: it's not checking the clock, just that it reponds to the status request

spec/support/command_helpers.rb
```diff
@@ -296,6 +296,40 @@ def set_date
...
+      { S0096: [
+        :year,
+        :month,
+        :day,
+        :hour,
+        :minute,
+        :second,
+      ] }
+    @site.log "Reset date", level: :test
```
emiltin, line 320: if resetting the clock is important to avoid other tests to fail, you should probably en ensure block to rest the clock even if some earlier part of the code fails

config/ruby.yaml
```diff
@@ -31,10 +31,15 @@ supervisor:
   status_response_timeout: 0.2
   status_update_timeout: 0.2
 
+scripts:
+  activate_alarm: /home/seguro/RSMP/scripts/activate_alarm.sh
```
emiltin, line 35: i don't like path to local folders in the repo. preferably. when testing against the ruby tlc, what does this script do?

spec/site/alarm_spec.rb
```diff
@@ -67,4 +67,43 @@ def check_scripts
   #    system(SCRIPT_PATHS['deactivate_alarm'])
   #  end
   #end
+
+  it 'Disruption test', :script, sxl: '>=1.0.7' do |example|
```
emiltin, line 71: test name should be more descriptive, maybe something like "buffers alarms during disconnects"

spec/site/alarm_spec.rb
```diff
@@ -67,4 +67,43 @@ def check_scripts
...
+    check_scripts
+    component = COMPONENT_CONFIG['detector_logic'].keys.first
+
+    TestSite.isolated do |task,supervisor,site|
```
emiltin, line 75: you should use TestSite.disconnected, and and place the system() call inside that block.
TestSite.isolated will first connect, when disconnect, which will add unnecessary delay to the test

spec/site/alarm_spec.rb
```diff
@@ -67,4 +67,43 @@ def check_scripts
...
+      start_time = Time.now
+      site.log "Waiting for alarm", level: :test
+      message, response = nil,nil
+      expect do
+        response = site.wait_for_alarm task, component: component, aCId: 'A0302',
+          aSp: 'Issue', aS: 'Active', timeout: RSMP_CONFIG['alarm_timeout']
+      end.to_not raise_error, "Did not receive alarm"
+
+      delay = Time.now - start_time
+      site.log "alarm confirmed after #{delay.to_i}s", level: :test
```
emiltin, line 93: try using log_confirmation() from log_helpers.rb to handle the time measurement and logging. see other tests for how it's used.

spec/site/alarm_spec.rb
```diff
@@ -67,4 +67,43 @@ def check_scripts
...
+      system(SCRIPT_PATHS['deactivate_alarm'])
+
+      alarm_time = Time.parse(response[:message].attributes["aTs"])
+      expect(alarm_time).to be_within(1.minute).of Time.now.utc
+      expect(response[:message].attributes['rvs']).to eq([{
+        "n":"detector","v":"1"},
```
emiltin, line 99: it's unfortunate that this values are not explicit when raising the alarm via the script. it should be clear when reading the test why these values are expected. can we perhaps call the script using ENV values to set the alarm values? @otterdahl?

emiltin (changes requested):

emiltin:
thanks for the PR! got a few inputs, se my review above

spec/site/alarm_spec.rb
otterdahl: Yes. Raising alarm scripts are very specific to the test environment and should preferably not be a part of the repo itself.

The path to a script could either be specified in a directly in workflow or specified using environment value. Each workflow file is specific for a test environment.

spec/site/alarm_spec.rb
```diff
@@ -67,4 +67,43 @@ def check_scripts
...
+    check_scripts
+    component = COMPONENT_CONFIG['detector_logic'].keys.first
+
+    TestSite.isolated do |task,supervisor,site|
+      site.log "Checking connection", level: :test
+      # Is an explicit check needed?
+    end
+      
+    # Deactivate alarm 
+    system(SCRIPT_PATHS['activate_alarm'])
```
otterdahl, line 81: The comment says "deactivate alarm", but the script says "activate alarm"?

spec/site/alarm_spec.rb
otterdahl: All those values are very specific to the alarm, but we can't be sure that all those are set in this specific way. (This is a problem with the alarm test that is currently in the master branch as well)

I think it is enough that we check:
1. The alarm is of type "A0302" is "Issue" and "Active"
2. The alarm received within the last 1 minute (not a buffered alarm)

The alarm code "A0302" could possibly be fetched from the workflow file or environment variable

emiltin: The files in config/ are meant to be templates. Based on these you create customized versions for the particular equipment that you test. The script paths should be in the customized config files.
We can either ignore this for the moment  and just keep our customized configs in the repo (i suggest so), or move them to a separate repo.

spec/site/alarm_spec.rb
emiltin: the ekstra delay will occur in case that site is already disconnected. in that case there is no need to connect and then disconnect. using TestSite.disconnected will see that it's already offline and just go ahead.

spec/support/command_helpers.rb
sweco-seguro: Cool, I'll move it there.

spec/support/command_helpers.rb
sweco-seguro: Smart, I'll do that.

spec/support/command_helpers.rb
sweco-seguro: The reseting of time part is done in multiple places though, so I made a helper function called reset_Date which does just that.

config/ruby.yaml
sweco-seguro: My bad, forgot to remove them before I committed. The script do nothing, I just added them to make sure 'check_scripts' didn't raise any errors when I was testing.

spec/site/alarm_spec.rb
sweco-seguro: Sure, sounds good to me! I'll change it.

spec/site/alarm_spec.rb
sweco-seguro: Right, I'll change it. I take it that no explicit connection check is needed as part of the test then?

spec/site/alarm_spec.rb
sweco-seguro: Lazy copypasting on my part, my bad.

spec/support/command_helpers.rb
sweco-seguro: I see, I'll make sure it actually checks the clock.

spec/site/alarm_spec.rb
sweco-seguro: TestSite.disconnected doesn't work unfortunately. See issue #67

sweco-seguro:
Thanks for the reviews! I've fixed most of the changes except the TestSite.disconnected one, see issue #67.

Should I push it even though it fails?

.gitignore
```diff
@@ -3,6 +3,7 @@ log
 .DS_Store
 config/secrets.yaml
 config/validator.yaml
+config/ruby_gunnar.yaml
```
emiltin, line 6: this should not be in the repo. better to add config/private/ to the .gitignore file, and move your config there

spec/site/command_spec.rb
```diff
@@ -211,7 +211,37 @@
...
+      message, result = @site.request_status @component, convert_status_list(status_list), collect: {
+        timeout: SUPERVISOR_CONFIG['status_update_timeout']
+      }
+      status = "S0096"
+      expect(result[{"sCI" => status, "n" => "year"}]["s"]).to be == "2020"
+      expect(result[{"sCI" => status, "n" => "month"}]["s"]).to be == "9"
+      expect(result[{"sCI" => status, "n" => "day"}]["s"]).to be == "29"
+      expect(result[{"sCI" => status, "n" => "hour"}]["s"]).to be == "17"
+      expect(result[{"sCI" => status, "n" => "minute"}]["s"]).to be == "29"
+      expect(result[{"sCI" => status, "n" => "second"}]["s"]).to be == "51"
```
emiltin, line 242: i don't think you can expect the second or minute to the the same as what you set, since the imte passes between setting and getting the clock

spec/support/command_helpers.rb
```diff
@@ -283,40 +283,10 @@ def set_date
...
-      securityCode: SECRETS['security_codes'][1],
-      year: Time.now.year,
-      month: Time.now.month,
-      day: Time.now.day,
-      hour: Time.now.hour,
-      minute: Time.now.min,
-      second: Time.now.sec
-    }
-    send_command_and_confirm @task, command_list, "intention to set date"
+    reset_date
```
emiltin, line 286: might use ensure block

.gitignore
sweco-seguro: Fixed.

spec/site/command_spec.rb
sweco-seguro: That makes sense, I didn't consider that time passes.

emiltin:
Once you resolve the conflict in ruby.yaml, you can merge this PR.

spec/site/command_spec.rb
emiltin: Maybe a more robust approach would be to simply diff the requested and the received time, and expect them to be within some configured range. That would verify the content of the minutes and seconds atttributes, which don't look at now, while taking into account network delays.

Something along the lines of:
```ruby
time_diff = (received - set).abs
expect( time_diff ).to be <= max_diff
```

PR #62, opened 2021-01-27
Fix issues with workflows

otterdahl:
- [x] Prevent the alarm test to run using the ruby TLC
- [x] Rename workflows
- [x] Merge separate jobs in the each workflow to single job
- [x] Properly isolate the ruby TLC test with the other simulators (use a different port)
- [x] Move config files used by CI to subdirectory (cI/)
- [x] Configure rsmp and sxl version in config/dynniq.yaml
- [x] Rename validation log


emiltin:
In the latest run, I see two tests fail due to timeouts. Most timeouts for the ruby TLC are currently set to just 0.2s, but that is a bit arbitrary as that is the lowest value that seemed to work on my local machine. If it's too low, you will start to see some tests timeout, depending on what else is going on your machine.

I suggest we increase the timeouts to 1s in ruby.yaml, as see if this makes tests pass reliably.

otterdahl:
In the last run we get a lot of `Command request be09311b was rejected: Component ID not recognized.` I believe this is because the tests again the ruby TLC is not properly isolated from the other simulators. The test of the ruby TLC may run on the same machine that performs the Swarco ITC-2 and Dynniq EC-2 tests - which, of course, may cause conflict - unless we change to a different port for the ruby TLC test

emiltin:
Yeah I was having the same suspicion. Yes, you can use to a different port for the Ruby TLC. This can easily be set in the config files.
Maybe we should add a folder config/ci in the rsmp gem and in rsmp_validator to keep configs we're using for the github ci?

emiltin:
Can we perhaps rename tests?
Dynniq EC-2
Swarco ITC-2
rsmp gem

otterdahl:
> Can we perhaps rename tests?
> Dynniq EC-2
> Swarco ITC-2
> rsmp gem

Sure

otterdahl:
> Yeah I was having the same suspicion. Yes, you can use to a different port for the Ruby TLC. This can easily be set in the config files.

I understand that  `port` in tlc.yml is used for the Ruby TLC. But where do I change the port which rspec uses?

> Maybe we should add a folder config/ci in the rsmp gem and in rsmp_validator to keep configs we're using for the github ci?

Yes. Perhaps. Every config file is very specific to a test environment. To move them to config/ci is fine by me

emiltin:
you set the port of the supervisor in ruby.yaml, e.g:
```yaml
supervisor:
  port: 13111
  ...
```

emiltin:
ec-2 has failures because it tries to run 1.0.13 tests.e.g:
RSMP site status S0025 time-of-green/time-of-red

We need to configure rsmp and sxl version in config/dynniq.yaml, like it's done in swarco.yaml:

```yaml
supervisor:
  sites:
    :any:
      rsmp_versions:
        - 3.1.2
      sxl_version: 1.0.7
...
```

emiltin:
EC-2 sends aggregated status with missing fS. I've opened an issue at https://github.com/rsmp-nordic/rsmp_implementations/issues/7

emiltin:
Currently the log artifacts generated by rsmp_validator are currently called:

ec2: validator.log
itsc2: validator.log
ruby: rsmp.log

I find this a bit confusing. When downloading logs it's also a bit hard to keep track of what they relate to. Could we perhaps use something like:

ec2: validator_ec2.log
itsc2: validator_itc2.log
ruby: validator_ruby.log


If it's easy, maybe we could also append a timestamp to the file name.

emiltin:
The name of the artifact is also different for the ruby test, as it's called 'rsmp log' in the Artifacts popup, whereas the other checks use 'rspec validation'. Maybe use 'validator log' for all three?

otterdahl:
> The name of the artifact is also different for the ruby test, as it's called 'rsmp log' in the Artifacts popup, whereas the other checks use 'rspec validation'. Maybe use 'validator log' for all three?

Actually, the rsmp log was the log file from ruby TLC, not the validation.log. The validation.log was missing. I've added it now

> ec2: validator_ec2.log
> itsc2: validator_itc2.log
> ruby: validator_ruby.log
> 
> If it's easy, maybe we could also append a timestamp to the file name.

Done. Hopefully it works

otterdahl:
> ec-2 has failures because it tries to run 1.0.13 tests.e.g:
> RSMP site status S0025 time-of-green/time-of-red
> 
> We need to configure rsmp and sxl version in config/dynniq.yaml, like it's done in swarco.yaml:
> 
> ```yaml
> supervisor:
>   sites:
>     :any:
>       rsmp_versions:
>         - 3.1.2
>       sxl_version: 1.0.7
> ...
> ```

Fixed (Hopefully)

otterdahl:
> Maybe we should add a folder config/ci in the rsmp gem and in rsmp_validator to keep configs we're using for the github ci?

Done.

emiltin:
cool.
do you know why all three checks are queued?

otterdahl:
> do you know why all three checks are queued?

I made many commits in a short amount of time. It seems each commit needs to be individually tested (takes approx 10 min) and they are stacked up in a queue

emiltin:
ok i see

emiltin:
looking good!
ruby passes all tests. ec2 is down to 27 failed, many due to 'not supported'. itc2 has 12 failed, which seem to be consistent.

otterdahl:
Yes, looks good. Perhaps we should merge this PL. We can open another one if needed. What do you think?

Issue #57, opened 2021-01-22
swarco simulator closes connection due to "no reply from server"


emiltin:
It closes at  13:53:39, stating 'No reply from server'. But what reply is it missing? I don't see anything that't not being acknowledged by the validator, and the last watchdog ws received 10 seconds earlier.

Simulator log:

```
(2021-01-22 13:53:29) : [INFO]	 RECEIVING : {"mType":"rSMsg","type":"Watchdog","wTs":"2021-01-22T13:53:29.628Z","mId":"be3b666a-6993-41eb-a709-7b8f04b12228"}

(2021-01-22 13:53:29) : [INFO]	 RECEIVING : {"mType":"rSMsg","type":"MessageAck","oMId":"7cd5c026-6471-41ef-9439-153c6c7c190f"}

(2021-01-22 13:53:29) : [INFO]	 SENDING : {"mType":"rSMsg","type":"MessageAck","oMId":"b2036e12-93e1-438f-9342-6c54f2b3b52d"}

(2021-01-22 13:53:29) : [INFO]	 Received Command Request
(2021-01-22 13:53:29) : [INFO]	 Controller Code : KK+AG9998=001TC000.
(2021-01-22 13:53:29) : [INFO]	 Checking cc-id M0006
(2021-01-22 13:53:29) : [INFO]	 Checking cc-id M0006
(2021-01-22 13:53:29) : [INFO]	 Checking cc-id M0006
(2021-01-22 13:53:29) : [INFO]	 SENDING : {"mType":"rSMsg","type":"MessageAck","oMId":"3e85db45-a8d7-499e-82b9-58d1a30a46ab"}

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0008

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0004

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0104

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0005

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0007

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0006

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0013

(2021-01-22 13:53:29) : [INFO]	 COMMAND SET: M0006, M0001

[...]

PR #63, opened 2021-01-27
Investigate duplicate connection and "no reply" issue with ITC-2

otterdahl:
Disable alarm for ITC-2 in order to investigate issue #56 and #57.

- Does it make the duplicate connection issue go away?
- Does it make the "No reply from server" issue go away?

### DIFF
--- a/.github/workflows/swarco_itc2.yml
+++ b/.github/workflows/swarco_itc2.yml
@@ -32,6 +32,8 @@ jobs:
       run: bundle config set path '~/rsmp_gems/bundle'
     - name: Run bundle install
       run: bundle install
+    - name: Restart rsmp service and clear buffer
+      run: killall rsmpd
     - name: Run tests
       run: bundle exec rspec spec/site
     - name: Rename validation.log
--- a/.github/workflows/swarco_itc2.yml
+++ b/.github/workflows/swarco_itc2.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Run bundle install
       run: bundle install
     - name: Run tests
-      run: bundle exec rspec spec/site
+      run: bundle exec rspec --tag '~script' spec/site
     - name: Rename validation.log
       if: always()
       run: mv log/validation.log log/validation_itc2_$(date +%F_%H-%M-%S).log
--- a/.github/workflows/swarco_itc2.yml
+++ b/.github/workflows/swarco_itc2.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Run bundle install
       run: bundle install
     - name: Run tests
-      run: bundle exec rspec --tag '~script' spec/site
+      run: bundle exec rspec spec/site
     - name: Rename validation.log
       if: always()
       run: mv log/validation.log log/validation_itc2_$(date +%F_%H-%M-%S).log


otterdahl:
Without the alarm test, the problems seems gone.

emiltin:
Does it try to open a separate connection when sending alarms? That would be a bug, I suppose.

otterdahl:
Let's re-enable the alarm test. But this time, do not trigger the alarm. Let it time out and see what happens. I've disabled the trigger in the ITC simulator

emiltin:
no double connections this time

emiltin:
how does it trigger the alarm? could it be the case that when you click on the gui, it somehow uses a separate process, causing a separate connection to be opened? Would be a bug.

otterdahl:
> how does it trigger the alarm? could it be the case that when you click on the gui, it somehow uses a separate process, causing a separate connection to be opened? Would be a bug.

It triggers the alarm by:
1. Temporarily change the internal parameter F12DN23 from 0 to 1 using a serial interface. This parameter represent the maximum time for a detector logic to be permanently "on" (active) before a detector fault is set. Value 0 disables this check
```bash
nc localhost $PORT >/dev/null << EOF
LOGIN USER -l 1 [removed]
LOGIN USER -l 2 [removed]
F12DN23=1
EXIT
EOF
```

2. Trigger detector 1 by simulate a mouse press on a specific coordinate in the GUI. It does this using the xdotool
```bash
# GUI coordinates for detector 1
DET_1_XY="121 579"

# Trigger detector 1 (=ON)
function trigger_detector {
        xdotool search --name "ITC Simulator" \
                windowactivate --sync %1 \
                mousemove --sync $DET_1_XY \
                click --clearmodifiers 1 \
                windowactivate $(xdotool getactivewindow)
```

3. Check that the detector has been activated using the serial interface

```bash
# Check status of detector 1 (1=ON, 0=OFF)
function check_detector {
        return `echo -e "LOGIN USER -l 1 [removed]\nLOGIN USER -l 2 [removed]\nDSTAT\n\033\nexit" |
                nc localhost $PORT |
                tail -n 1 |
                awk 'BEGIN { FS = "" } ; { print $6 }'`
}
```
4. Repeat until we're sure that the detector is active
```bash
check_detector
while [ $? -eq 0 ]
do
        trigger_detector
        check_detector
done
```

otterdahl:
The trigger script is sometimes very slow due to `xdotool`. I wonder if speed is a factor? 

I've tried to remove some checks to the script. I'll try running it again

emiltin:
I think we should file double comnnections as a bug in the simulator. No matter what, it's incorrect to connect twice to the same serverl using the same site it.

otterdahl:
> I've tried to remove some checks in the script. I'll try running it again

Without the check. I didn't trigger the alarm at all, so re-enable it again

otterdahl:
I'll try launching to simulator using non-root credentials (using fakechroot rather then chroot). This would allow us to restart the simulator and clear the buffer before each run. I wonder if this changes the outcome.

otterdahl:
Restarting the rsmp service and clearing the buffer has no effect.

otterdahl:
Anyway. I'm closing this PL. It seems to be triggered by the alarm, but I can't find the cause.

PR #64, opened 2021-01-29
M0006 and M0008: Wait until status switches

otterdahl:
Fixes issue where the test only checked a single time for the expected outcome, but these command might take a couple of seconds to complete.

- [x] M0006 works as expected
- [x] M0008 works as expected


otterdahl:
Looks good. The tests now works as expected. Two fewer failures in the Swarco simulator

Issue #70, opened 2021-02-04
test timestamps when using M0104


emiltin:
I added support in the ruby tlc for M0104, setting the clock. But I noticed that messages like AggregatedStatus and CommandResponse includes a timestamp that is still using the system clock. But we don't yet test these timestamps, we only check what S0096 returns.

_Originally posted by @emiltin in https://github.com/rsmp-nordic/rsmp_validator/issues/69#issuecomment-772439933_

emiltin:
Theses messages include timestamps:

- AggregatedStatus
- Alarm
- StatusResponse
- CommandResponse
- Watchdog

GunnarRolander:
I'll take a look at this.

GunnarRolander:
In the pull request it checks StatusRequests, should we make tests to check the rest of the message types?

emiltin:
yes. you might even want to move the statusresponse timestamp check to a separate test

sweco-seguro:
Should I maybe create a new test file even? Seeing as we're sending commands, statusRequests, and probably will be triggering alarms too. Or do we still consider it a test of M0104 and should keep it in command_spec?

Btw, is there any way to request an aggregated status? I can't seem to find any.

otterdahl:
> Btw, is there any way to request an aggregated status? I can't seem to find any.

It's only possible with RSMP 3.1.5, see [Aggregated status request message](https://rsmp-nordic.github.io/rsmp_core/applicability/basic_structure.html#aggregated-status-request-message)

emiltin:
There is no support yet, but we should add that. Tracking in https://github.com/rsmp-nordic/rsmp/issues/18

emiltin:
I suggest adding a new file spec/site/clock_spec.rb for all the tests related to m0104 and s0096.
We're planning a reorganization anyways, see #22

emiltin:
Support for aggregated status request has been added in the gem, and there's a test in aggreated_status_request_spec.rb that tests it's use. Note that the test is tagged with rsmp>=3.1.5, meaning it will be filtered out unless the config for the tested equipment meets this criteria, see config/ci/rsmp_gem.yaml for an example. EC-2 and ITC-2 of course does not yet support 3.1.5 or aggregated status request.

emiltin:
done

PR #71, opened 2021-02-09
time synch now checks message timestamp

GunnarRolander:
Time synch tests now check StatusRequest timestamps as mentioned in #70

### DIFF
--- a/spec/support/command_helpers.rb
+++ b/spec/support/command_helpers.rb
@@ -1,12 +1,12 @@
 module CommandHelpers
   def send_command_and_confirm parent_task, command_list, message, component=@component
-    request, response = nil, nil
+    result = nil
     log_confirmation message do
-      request, response = @site.send_command component, command_list, collect: {
+      result = @site.send_command component, command_list, collect: {
           timeout: SUPERVISOR_CONFIG['command_response_timeout']
         }
     end
-    [request, response]
+    return *result   # use splat '*' operator
   end
 
   def build_command_list command_code_id, command_name, values
--- a/spec/support/command_helpers.rb
+++ b/spec/support/command_helpers.rb
@@ -271,18 +271,20 @@ module CommandHelpers
     send_command_and_confirm @task, command_list, "intention to set security code"
   end
 
-  def set_date
+  def set_date date
     @site.log "Set date", level: :test
     command_list = build_command_list :M0104, :setDate, {
       securityCode: SECRETS['security_codes'][1],
-      year: 2020,
-      month: 9,
-      day: 29,
-      hour: 17,
-      minute: 29,
-      second: 51
+      year: date.year,
+      month: date.month,
+      day: date.day,
+      hour: date.hour,
+      minute: date.min,
+      second: date.sec
     }
-    send_command_and_confirm @task, command_list, "intention to set date"
+
+    message, result = send_command_and_confirm @task, command_list, "intention to set date"
+    return message, result
   ensure
     reset_date
   end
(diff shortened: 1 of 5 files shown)


emiltin:
reset_date() in command_helpers.rb must use UTC time. RSMP uses UTC for all timestamps.

There's might also be an issue with using in inside an ensure block, since it involves expect check, and might therefore raise exceptions.

sweco-seguro:
What is up with this error?
´´´RSMP::NotReady:
       RSMP::NotReady´´´

I get it quite often when running against the rsmp gem TLC.

spec/support/command_helpers.rb
```diff
@@ -271,18 +271,20 @@ def set_security_code level
...
+      month: date.month,
+      day: date.day,
+      hour: date.hour,
+      minute: date.min,
+      second: date.sec
     }
-    send_command_and_confirm @task, command_list, "intention to set date"
+
+    message, result = send_command_and_confirm @task, command_list, "intention to set date"
+    return message, result
```
emiltin, line 287: the two lines can be reduced to just:
send_command_and_confirm @task, command_list, "intention to set date"

spec/support/command_helpers.rb
```diff
@@ -291,12 +293,12 @@ def reset_date
     @site.log "Reset date", level: :test
     command_list = build_command_list :M0104, :setDate, {
       securityCode: SECRETS['security_codes'][1],
-      year: Time.now.year,
-      month: Time.now.month,
-      day: Time.now.day,
-      hour: Time.now.hour,
-      minute: Time.now.min,
-      second: Time.now.sec
+      year: Time.now.utc.year,
```
emiltin, line 296: better to use Time.now.utc once and store in a local, then access year, month, etc in that.

emiltin:
>      RSMP::NotReady:

It means the site did not complete the connction sequence within the expected time.
You can try to increase the ready_timeout in config/rsmp_gem.yaml. It's currently set to 1s which works on my machine, but maybe you need a higher value. Try setting it to 10s just to see if that is the cause of the problem.

emiltin:
I was to quick about NotReady. In fact it's raised by methods like request_status() if the proxy is not ready, ie. there is no connection yet or the handshake is not complete yet. Not sure why you get it, as each time you use e.g. TestSite.connection, it should wait for ready state in this bit:
```ruby
    @remote_site.wait_for_state :ready, RSMP_CONFIG['ready_timeout']
```

emiltin:
When there are changes in master, I like to rebase on master, rather then merge. ie:
```
git rebase master
```
That will replay your commits on top of the latest master. You commits will be kept together and there will be no merge messages in the commit history.

emiltin:
Hi Hunnar, I've pushed some changed.
There's a few tricky issues.
a) request_status() returns [request, result]. Requst is the message send. Result is not a message, but a hash with the collected status values. This is because the response can be split into multiple messages. This means that if we want to look at the RSMP timestamp, we need an actual response messge, and the collect() methods don't all return that currently. No big deal, since we collect it outself 'Check status response timestamp', but I will consider making it easier to get to the messages from the collect methods.

b) response.timestamp is an internal timestamp which is set when the mesages is received, using the system clock on the local machine. The RSMP timestamp is in response.attributes['sTs'] for status responses, but the name varies (enfortunetely) between messages.

c) the date cannot be reset using an ensure block in the set_date() helpers, otherwise the date will be reset before we can do our checks. I added a small helper with_date_sets with handles the resetting.

d) i've uddated the rsmp gem. it now uses the time set with m0104 for the various mesage timestamp. (i think watchdogs might be missing). all the tests currently pass

emiltin:
I added that last to two tests. the gem will now return the actual rsmp messages when collecting, in addition to the attribute values.

There's some aspects of how excetions are handles which is a bit akward when returning multiple values, so i might do another refactoring of the collect methods at some point.

But for now, it's a bit easier to check the content of actual rsmp messages collected. It also means we don't need to do any manual collection (and task handling) in these clock specs.

PR #76, opened 2021-02-22
try fixed version of json_schemer

emiltin:
Let's see if the offical fix works. The fix has not yet been published as an updated gem, so this references a specific commit.


emiltin:
It works

otterdahl:
If it works then I guess we can merge this PL?

emiltin:
merged

PR #81, opened 2021-03-09
Yard

emiltin:
Using Yard to document rspec tests.
Picking up rspec specs require yard extensions, which are a bit tricky to write. So far, the generated docs will show a list of Features, but more work is needed.


emiltin:
@otterdahl we could use a github action to run yardoc when pushing, and .gitignoring /doc? Otherwise we run yardoc manually and commit the docs to the repo.

otterdahl:
> @otterdahl we could use a github action to run yardoc when pushing, and .gitignoring /doc? Otherwise we run yardoc manually and commit the docs to the repo. 

Yes. I believe it is possible. Not sure how to commit to /doc using a script
though.

emiltin:
I don't think actions should commit to the repo. We could publish the docs somewhere, like github pages. But we would want PR branches to do that, so let's consider it a bit more before doing anything. Anyways first step is getting the yard extensions more complete.

emiltin:
@GunnarRolander you're welcome to add commits to this branch if want to move descriptions into the code

sweco-seguro:
Great! Do you mind if I rebase it to get the latest time synch test etc?

emiltin:
yes you can rebase it on master

emiltin:
I worked on the extensions and output generation.

The branch was rebased on master. If you have local commit, rebase you local yard branch on this branch, eg:
```
# on yard branch
% git fetch origin
% git rebase origin/yard
```

emiltin:
We also need to improve out use or RSpec context, because they are used to group specifications in the documentation.

It's what you specify wit the RSpec describe() call:
```ruby
RSpec.describe 'RSMP site commands' do  
```

They can also be nested.

emiltin:
Hi Gunnar, I think there's a problem with your rebase. The file yard/yard_extensions.rb was added, but this file was previously renamed to yard/extensions.rb. And .yardopt needs to point to the renamed file, not the old.

emiltin:
Gunnar,
To make list works, you need to use markdown, ie. dots, not colons:
```
1. Given the site is connected
2. Request status
3. Expect status response before timeout
```

=>

1. Given the site is connected
2. Request status
3. Expect status response before timeout

I've comitted an example in the S0001 tests.

emiltin:
I think the first step "Verify connection" would perhaps be better described as a given, e.g. "Given the site is connected", as it's really more of a preparation, that the actual test. This mirrors the language used in Cucumber.

I also think it would be nice with a line first, describing in plain text what the test does. eg:

 ```
Verify that the controller responds to S0001.
 
1. Given the site is connected
2. Request status
3. Expect status response before timeout
```

**Comment by emiltin** [comment]
Nested RSpec context are now shown in the docs. Eg this spec organization

```ruby
RSpec.describe "Traffic Light Controller" do
  include StatusHelpers
  
  describe "Signal Plans" do

    # Verify that the controller responds to S0001.
    #
    # 1. Given the site is connected
    # 2. Request status
    # 3. Expect status response before timeout

    it 'responds to S0001 signal group status', sxl: '>=1.0.7' do |example|
      request_status_and_confirm "signal group status",
        { S0001: [:signalgroupstatus, :cyclecounter, :basecyclecounter, :stage] }
    end
  end
end
```

It will be shown in the docs with nested sections.

Source code can now also be toggled on/of for specs.

sweco-seguro:
Are all the current site tests related to Traffic Light Controllers, or are some more generic to different equipment? I.e. should all of them be nested inside a
```ruby
RSpec.describe "Traffic Light Controller" do
```

or should some, like disconnect, clock sync etc, be in some more general desciprtions?

emiltin:
Some tests would also be relevant to other types of equipment. I think we  should place them in a RSpec context called e.g. "Core". We should probably also use folders to separate core and tlc specs:
```
spec
  site
    core
    tlc
  supervisor
```

emiltin:
Note that all statuses and commads are we currently use are specific to traffic light controllers, since they're defined in the tlc sxl. This includes M0104 and S0096. subscribe_spec.rb also uses S0001 which is defined in the tlc sxl.
None of the tests is clock_spec.rb can therefore be said to belong to Core.

emiltin:
@GunnarRolander let me know when you've completed import of test descriptions. Then I'll look at merging.

sweco-seguro:
Alright, all the documentation from the test list excel is now in the code. I updated it where necessary and added some documentation to aggregated status and disconnect too.

spec/site/connect_spec.rb
```diff
@@ -51,22 +51,52 @@ def check_sequence version
     end
   end
 
+# Verify that the handshake sequence upon connection is correct for 3.1.1
+#
+# 1. Given the site is new, connected and version 3.1.1
```
emiltin, line 56: what is the meaning of "new"?

spec/site/connect_spec.rb
```diff
@@ -51,22 +51,52 @@ def check_sequence version
...
+# 2. Send and receive handshake messages
+# 3. Expect the handshake messages to be in the specified sequence corresponding to version 3.1.1
+# 4. Expect the site to be ready
```
emiltin, line 59: the 'site' object is a local proxy for the real equipment.
what we expect is that the site proxy object reaches the ready state. this means that the connection sequence is complete. i suggest "expect the connection sequence to be complete"

spec/site/connect_spec.rb
```diff
@@ -51,22 +51,52 @@ def check_sequence version
...
   it 'exchanges correct connection sequence of rsmp version 3.1.1', rsmp: '3.1.1' do |example|
     check_sequence '3.1.1'
   end
 
+# Verify that the handshake sequence upon connection is correct for 3.1.2
+#
+# 1. Given the site is new, connected and version 3.1.2
```
emiltin, line 66: maybe "Verify the connection sequence when using rsmp core 3.1.2"

emiltin, line 66: maybe "Given the site is connected and using core 3.1.2"

spec/site/connect_spec.rb
sweco-seguro: I was thinking of the site object, and couldn't come up with anything better at the time.

spec/site/connect_spec.rb
sweco-seguro: Right, that makes sense. I'll change it.

spec/site/connect_spec.rb
sweco-seguro: Sounds good, changed it.

sweco-seguro: Sounds good, changed it.

spec/site/connect_spec.rb
```diff
@@ -1,14 +1,14 @@
 RSpec.describe "RSMP site connection" do
 
-  def check_connection_sequence version, expected
+  def check_connection_sequence version, Expected
```
emiltin, line 3: what's up with the capitalizatin here?

sweco-seguro: Woops, I managed to search-replace a tad more than I aimed for. Fixed it.

emiltin: alright :-)

emiltin:
@GunnarRolander is this ready to merge, or do you want to move add more documentation in this first round?

sweco-seguro:
I think it's ready to merge. Everything in the excel and some extra has been added.

Issue #22, opened 2020-03-16
refactor status_spec.rb file


emiltin:
the file is geting too long.

many of the tests do nothing more than verify that a response is received. it then verifies the format of the response, but can be removed, this the format is already verified by the JSON schema. we just need to make sure any patterns matching is also in the schema.

tests that just verifies that a respons is received are very similar and could be refactored to use a commend helper method, reducing the amount of code.

of course we might also think of way to do additional verification, specific to each status.

otterdahl:
True. Many of the tests only verfies that the response is recieved. This is different from the **command_spec.rb** where each command typically is verified against a status. The level of supported statuses differs much between the manufacturers.

emiltin:
Patterns matching of comma separated lists are not working as intended. Eg.```/^True|False(,True|False)*$/``` matches anything starting with "True", like True,xxx", but does not match ""False,False". But as mentioned I suggest moving these regex patterns to the schema and remove from the ruby test file.

emiltin:
I suggest we reorganize test files according to functionality instead of status/command/alarm. Testing commands will typically involve requesting statuses, etc. So we could organize by e.g. signal plan, inputs, etc instead, and then use commands, statuses, alarms, etc. as needed to test that specific functionality.

otterdahl:
OK. And while reorganizing, also differentiate generic RSMP tests to the specific TLC tests

emiltin:
factoring out common functionality to helpers has been implemented

emiltin:
done

PR #82, opened 2021-03-19
reorganize tests in functional areas

emiltin:


emiltin:
fixes #22
hi @otterdahl what do you think of this way of organizing tests? It's also reflected in the yard documentation structure.

otterdahl:
I have only looked at it briefly. I think it is a good effort. But my initial thoughts on organization are:
* It would be logical to move TLC specific tests to `spec/site/tlc/` so that we could add e.g. `spec/site/vms/` in the future.
* I also think that is good to keep the test cases in consecutive order, e.g. A001, A002, A003 and so on (where possible), since this is the way you normally [read the SXL](https://rsmp-nordic.github.io/rsmp_sxl_traffic_lights/sxl_traffic_light_controller.html#alarms)
* It is good to separate functional and non-functional tests. For instance, tests that only checks if S0001, S0002 (and so on..) can be requested and the response is valid - and tests that checks a command and status combination, e.g. M0002 (set time plan) and S0014 (get time plan)

emiltin:
Created spec/site/tlc as suggested and also spec/site/core.
I think test are sorted by status/command/alarm code, within rspec each group (functional area).

Regarding functional/not: If we organize tests by functional areas (signal plans, clock, detector logics) etc, then there will be both types of tests within each of these topics. But maybe we can tag tests, so you can run e.g all a quick scan of what statuses it responds to?

Or we can add a folder called like spec/site/tlc/scan and move all the tests that simply check if the site responds into that folder?

otterdahl:
> Regarding functional/not: If we organize tests by functional areas (signal plans, clock, detector logics) etc, then there will be both types of tests within each of these topics. But maybe we can tag tests, so you can run e.g all a quick scan of what statuses it responds to?
> 
> Or we can add a folder called like spec/site/tlc/scan and move all the tests that simply check if the site responds into that folder?

I think both approaches are OK.

I think there is also a need for a category for stress tests, e.g. fill the communication buffer, send random data, etc. But let's test one of these approaches and see if it needs to changed in the future.

emiltin:
alright, then I suggest we merge this. then we have organisation by topic. If/when we need to run e.g all status tests, we can add tags or another folder.

PR #83, opened 2021-03-23
Changed documentation to use Given - When - Then

GunnarRolander:
I've changed the documentation to use Given, When and Then instead.

I'm not 100% sure about some of the texts.


emiltin:
Hi,
I'm not sure about this. I like the more standardized language, but I'm worried that we're loosing useful details about how the tests work. If we end up with things like "try A, then expect A to work", it's too general.

@weco-sejnje

PR #89, opened 2021-05-19
support testing different rsmp core/sxl versions

emiltin:
The new gem 'rsmp_schemer' handles validating different version of the core/sxl json schema.

Todo:
handle behavioural differences between core versions

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp', :github => 'rsmp-nordic/rsmp', :ref => 'd144a72', :submodules => true
+gem 'rsmp', :github => 'rsmp-nordic/rsmp', :ref => '23bca01', :submodules => true
 gem 'rspec'
 gem 'activesupport'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp', :github => 'rsmp-nordic/rsmp', :ref => '08789e8', :submodules => true
+gem 'rsmp', :submodules => true
 gem 'rspec'
 gem 'activesupport'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp', :github => 'rsmp-nordic/rsmp', :ref => 'd6094e6', :submodules => true
+gem 'rsmp', :github => 'rsmp-nordic/rsmp', :ref => '08789e8', :submodules => true
 gem 'rspec'
 gem 'activesupport'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp', :github => 'rsmp-nordic/rsmp', :ref => '23bca01', :submodules => true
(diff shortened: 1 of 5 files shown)


emiltin:
Hi @otterdahl, what's the status on the TLC integrations at the moment?

A few things need to change in the configs, which iIcannot do, because they are gitignored and local to the test setup:


- the :any field in supervisor.yaml must be renamed to :unknown
- rsmp_gem.yaml (and similar for the equipments) must specify the sxl, e.g.:
```yaml
sxl: tlc
sxl_version: 1.0.15
```

otterdahl:
The current status of the TLC:
TLC-2

* EC-2: Connected, but need configuration in order to test all features of RSMP.
  * We need to ask Dynniq to make config
* ITC-2: Works, but issues with simulator makes it difficult to test all features. Issues with alarms
* ITC-3: Needs updated config in order to test all features of RSMP. Issues with alarm message.
  * We've asked Swarco Sweden to make a new config
  * We've asked Swarco Tech to fix the alarm message
  * Not currently connected due to windows update of the laptop connected to the TLC. The Github runner needs to be auto-started, needs investigation

emiltin:
Thank you for the status

emiltin:
ITS-2 fail with:
```
Run killall rsmpd
rsmpd: no process found
Error: Process completed with exit code 1.
```

Do you know why that might be?

emiltin:
For now, 'tlc' will be used as the default if no SXL is specified in the equipment config.

otterdahl:
Problem with ITC-2 should be fixed now

emiltin:
I noted that EC-2 has a step called 'Setup Ruby', but ITC-2 does not have this.

emiltin:
ITC-2 still bombs out:
Run killall rsmpd
rsmpd: no process found

otterdahl:
Yes. The EC-2 & ITC-3 runs on windows and ruby is set up every time the test is run. This is the preferred method. The ITC-2 simulator on the other hand is running linux and uses the installed version of ruby

otterdahl:
Ok. Fixed (hopefully). Try again now.

emiltin:
What this branch does it that it uses the core version negotiated during connection to select the JSON schema used for validating messages. The SXL is a bit odd, since the site only sends the version, not the type. So the type comes from the site config on the supervisor side, while the version is from the negotiation. This is then also used to pick the scheme used for validation. The rsmp_schemer gem handles the validation, and uses git submodules to keep each all the different schema version checked out.

emiltin:
Still problems:
Run killall rsmpd
rsmpd(1292721): Operation not permitted
rsmpd: no process found
Error: Process completed with exit code 1.

otterdahl:
I'm having problems with the hypervisor running the ITC-2 virtual machine:
```
[   28.373298] sr 0:0:0:0: [sr0] tag#22 unaligned transfer
[   28.373299] blk_update_request: I/O error, dev sr0, sector 6 op 0x0:(READ) flags 0x0 phys_seg 1 prio class 0
[   28.373300] Buffer I/O error on dev sr0, logical block 6, async page read
```

I need the ITC-2 to be off until this is fixed

emiltin:
ok

emiltin:
All EC-2 tests fail because the site doesn't connect:

```
RSMP::TimeoutError:
  Site 'any'' did not connect within 60s
```

PR #91, opened 2021-05-21
Yellow flash

emiltin:
Demonstrates how to write a test that uses S0001 to check that groups go to a specific state.

In this case, we check that all groups go to 'c', which means yellow flash.
Afterwards we check we turn off yellow flash and check that groups are NOT 'c'.

### DIFF
--- a/spec/site/tlc/yellow_flash.rb
+++ b/spec/site/tlc/yellow_flash.rb
@@ -1,0 +1,23 @@
+RSpec.describe "Traffic Light Controller" do
+  include StatusHelpers
+  include CommandHelpers
+
+  describe "Yellow Flash" do
+
+    it 'M0001 set yellow flash', sxl: '>=1.0.7' do |example|
+      Validator::Site.connected do |task,supervisor,site|
+        prepare task, site
+
+        timeout =  10
+
+        switch_yellow_flash
+        wait_for_groups 'c', timeout: timeout      # c mean s yellow flash
+
+        switch_normal_control
+        wait_for_groups '[^c]', timeout: timeout   # not c, ie. not yellow flash
+      end
+    end
+
+  end
+end
+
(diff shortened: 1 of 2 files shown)


emiltin:
The test looks like this:
```ruby
switch_yellow_flash
wait_for_groups 'c', timeout: timeout      # c mean s yellow flash

switch_normal_control
wait_for_groups '[^c]', timeout: timeout   # not c, ie. not yellow flash
```


It uses a small helper function:
```ruby
def wait_for_groups state, timeout:
  timeout = 10
  regex = /^#{state}+$/
  wait_for_status(@task,
    "Wait for all groups to go to yellow flash",
    [{'sCI'=>'S0001','n'=>'signalgroupstatus','s'=>regex}],
    update_rate: 0,
    timeout: timeout
  )
end
```

This again uses lower-level helpers, but essentially, you can subscribe to a status, and wait until you receive a response that mathches a regular expression. In this case all groups should either be 'c', or NOT be 'c'.

emiltin:
@otterdahl do we want to merge this more functional test?

otterdahl:
> @otterdahl do we want to merge this more functional test?

It looks good, but it is hard to verify it:
* ITC-2 fails due to invalid alarm error
* ITC-3 seems offline (I've sent a email to Swarco)
* EC-2 won't connect, but it is online since I can reach the web interface

otterdahl:
This PL needs to be rebased on master

emiltin:
rebased on master

emiltin:
@otterdahl want this test merged? it was meant as an illustration of how to check that groups go to a specific state, but might be useful as a test case?

otterdahl:
Yes. I think it is good. We need more tests based on S0001 and checking that groups goes to a specific state. This is a good start.

PR #101, opened 2021-06-24
Run ITC-2 vm in a clean working environment

otterdahl:
In order to minimize the risk of letting the previous run affect future runs, don't run the TLC continuously.  Start the TLC vm each run and shutdown the vm when done.

- [x] Move the Github runner to the hypervisor instead
- [x] Make sure the hypervisor can run ruby correctly
- [x] Start the TLC vm
- [x] Tune any timeouts (it takes a minute or two to start the vm)
- [x] Set the time correctly
- [x] Make sure the TLC vm connects to the hypervisor
- [x] Make sure log files are fetched correctly
- [x] Stop the TLC vm

### DIFF
--- a/.github/workflows/swarco_itc2.yml
+++ b/.github/workflows/swarco_itc2.yml
@@ -36,6 +36,7 @@ jobs:
       if: always()
       run: mv log/validation.log log/validation_itc2_$(date +%F_%H-%M-%S).log
     - name: Stop ITC-2 simulator
+      if: always()
       run: virsh shutdown ts1001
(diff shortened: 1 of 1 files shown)


emiltin:
sounds like a good idea!
i noticed that there's a workflow step that waits 30s for the simulator to be ready. Can this be improve to continue as soon as it's ready, e.g. by polling a port once per second?

otterdahl:
Sure. It can be improved. The temporary 30s delay seems to work for now and it quite spot on.

Issue #42, opened 2020-12-03
Test case: Check startup interval


otterdahl:
Add a functional test when starting the controller.

The signal group status should begin with this sequence:
* Start-up interval 1: minimum 3 seconds
* Start-up interval 2: minimum 3 seconds
* Start-up interval 3: minimum 3 seconds

In Sweden, this is defined by Transportstyrelsens författningssamling [TSFS](https://www.transportstyrelsen.se/TSFS/TSFS%202014_30k.pdf), chapter 2, §17 (page 4)

otterdahl:
Test cases for start-up has been added, but time minimum durations needs to be verified

emiltin:
> Test cases for start-up has been added, but time minimum durations needs to be verified

This would be an enhancement of the existing checks for startup sequence?

otterdahl:
yes

emiltin:
implemented. if we need to improve the tests, we can open new issues

PR #104, opened 2021-06-28
ITC-2 CI: Improve start of VM check

otterdahl:
Improve the check of VM startup

* Check that the terminal port is open (part of the simulator)
* Check that the TLC completes the startup sequence. Sending commands during startup sequence may cause issues (perhaps write new test cases for this?)
* Timeout after 5 minutes

### DIFF
--- a/.github/workflows/swarco_itc2.yml
+++ b/.github/workflows/swarco_itc2.yml
@@ -26,8 +26,9 @@ jobs:
       run: bundle install
     - name: Start ITC-2 simulator
       run: virsh -c qemu:///system start ts1001
-    - name: Wait for ITC-2 simulator to start
-      run: sleep 30s
+    - name: Wait for ITC-2 simulator to start and completes startup sequence
+      timeout-minutes: 5
+      run: until `echo -e "ITC\nEXIT" | nc 192.168.103.205 10002 | grep -q -a "3-light"`; do echo Waiting...; sleep 5; done
     - name: Set the time of the controller
       run: ssh 192.168.103.205 ./set_date.sh
     - name: Run tests


emiltin:
Existing test that use Validator::Site.reconnected will close and reopen the connection to the site before running the actual test, and therefore implicitly test what happens when we send messages to the site right after connection.
But maybe we can write tests that more explicitly try to test what happens when we e.g. change signal plan during start up.
Do you have more concrete examples of what causes problems?

.github/workflows/swarco_itc2.yml
```diff
@@ -26,8 +26,9 @@ jobs:
       run: bundle install
     - name: Start ITC-2 simulator
       run: virsh -c qemu:///system start ts1001
-    - name: Wait for ITC-2 simulator to start
-      run: sleep 30s
```
emiltin, line 30: it would be great to replace the hardcoded 30s wait with something that waits until it's ready (but with a timeout)

otterdahl:
Well. I assumed that there is issues with sending commands during startup sequence. But now when I compare the outcome of this pull request I can't find any difference with the master branch.

But I find it strange that "M0001 set dark mode" and "M0008 activate detector logic" fails. I think they used to work?

emiltin:
The helper method wait_for_normal_control waits for three statuses:

```ruby
  def wait_normal_control
    # Wait for:
    # 'switched on' to be true (dark mode false)
    #  yellow flash status to be false
    # for startup mode to be false
    wait_for_status(@task,
      "dark mode off, yellow flash off, start-up mode off",
      [
        {'sCI'=>'S0007','n'=>'status','s'=>/^True(,True)*$/},
        {'sCI'=>'S0011','n'=>'status','s'=>/^False(,False)*$/},
        {'sCI'=>'S0005','n'=>'status','s'=>'False'}
      ]
    )
  end
```

But it looks like the validator only subscribes to S0007:
```
2021-06-28T14:34:22.684Z 41716    KK+AG9998=001TC000            Confirming switch to dark mode
2021-06-28T14:34:22.689Z 41716    KK+AG9998=001TC000 <--  7796  Sent StatusSubscribe {"mType":"rSMsg","type":"StatusSubscribe","ntsOId":"","xNId":"","cId":"KK+AG9998=001TC000","sS":[{"sCI":"S0007","n":"status","uRt":"0"}],"mId":"77969adc-51dd-4485-97a8-65852b212caa"}
2021-06-28T14:34:22.692Z 41716    KK+AG9998=001TC000 -->        Received MessageAck for StatusSubscribe 7796 {"mType":"rSMsg","type":"MessageAck","oMId":"77969adc-51dd-4485-97a8-65852b212caa"}
```

otterdahl:
> Existing test that use Validator::Site.reconnected will close and reopen the connection to the site before running the actual test, and therefore implicitly test what happens when we send messages to the site right after connection.

Yes. But when I wrote "startup sequence" I didn't mean the RSMP handshake, but the TLC startup interval. (see https://github.com/rsmp-nordic/rsmp_validator/issues/42). In Sweden it is:
1. At least 3 seconds yellow flash
2. At least 3 seconds yellow (red for pedestrians/cyclists)
3. At least 4 seconds red
After this interval the TLC proceeds with normal control

I've noticed that sometimes when sending commands during the startup interval, the TLC might not complete the command properly. The startup-mode can be checked with S0005 and is checked with wait_for_normal_control()

emiltin:
I understand that you mean after connection, but while the TLC is still running through it's TLC-related startup sequence.
We should be able to test sending commands during the tlc-startup sequence by restarting the TLC?

By comment about `wait_for_normal_control` above was incorrect. The test seems to work correctly. The test "M0001 set dark mode" fails simply because the TLC never reports that it goes to dark mode within the timeout.

otterdahl:
> We should be able to test sending commands during the tlc-startup sequence by restarting the TLC?

Yes. That would be a good test case

emiltin:
can this issue be closed or is there more to do?

otterdahl:
Yes. I think so. The only thing this pull request does it that it replaces the hardcoded 30s wait with a check for open port at 10002 on vm which indicates that the tic simulator has started. But perhaps this is unnecessary since the rspec tests waits for the incoming RSMP connection from the TLC anyway.

otterdahl:
> But perhaps this is unnecessary since the rspec tests waits for the incoming RSMP connection from the TLC anyway.

My mistake. If the vm hasn't started after 30 seconds the next step of the workflow fails (setting time of the tic), so there it is a point to verify that the the tic has started before moving to the next step in the workflow.

emiltin:
you replaced the 30s wait with a check for when a port is open?

merge at will

otterdahl:
Ok. Merging

PR #110, opened 2021-07-13
Added new test cases

sweco-seguro:
Addd new testcases based on Johan and Martins test list.

### DIFF
--- a/spec/site/tlc/system_spec.rb
+++ b/spec/site/tlc/system_spec.rb
@@ -87,7 +87,7 @@ RSpec.describe 'Site::Traffic Light Controller' do
       Validator::Site.isolated do |task,supervisor,site|
         prepare task, site
         site.wait_for_state :ready, Validator.config['timeouts']['ready']
-        wait_normal_control_and_status
+        verify_startup_sequence
       end
     end
 
--- a/spec/site/tlc/system_spec.rb
+++ b/spec/site/tlc/system_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe 'Site::Traffic Light Controller' do
     # 9. Wait for S0001 status "ffffffff"
     # 10. Wait for S0001 status "gggggg"
     # 11. Wait for S0020 status "control" 
-    it 'M0004 restart and added status check', sxl: '>=1.0.7' do |example|
+    it 'goes through startup sequence after M0004 restart', sxl: '>=1.0.7' do |example|
       Validator::Site.isolated do |task,supervisor,site|
         prepare task, site
         #if ask_user site, "Going to restart controller. Press enter when ready or 's' to skip:"
--- a/spec/site/tlc/system_spec.rb
+++ b/spec/site/tlc/system_spec.rb
@@ -60,6 +60,37 @@ RSpec.describe 'Site::Traffic Light Controller' do
       end
     end
 
+    # 1. Verify connection in isolated_mode
+    # 2. Send the control command to restart, include security_code
+    # 3. Wait for status response= stopped
+    # 4. Reconnect as Isolated_mode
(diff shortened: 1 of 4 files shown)


spec/support/command_helpers.rb
```diff
@@ -373,6 +373,14 @@ def wait_normal_control
     )
   end
 
+  def wait_normal_control_and_status
+    wait_for_status(@task,"controlmode startup", [{'sCI'=>'S0020','n'=>'controlmode','s'=>'startup'}])
+    wait_for_status(@task,"signalmode eeeee", [{'sCI'=>'S0001','n'=>'signalgroupstatus','s'=>'eeeee'}])
+    wait_for_status(@task,"signalmode ffffffff", [{'sCI'=>'S0001','n'=>'signalgroupstatus','s'=>'ffffffff'}])
```
emiltin, line 379: what we expect is a sequence of modes like e,f,g. wait_for_status() will not complain if other modes are seen first. so a sequence like e,f,e,g will be accepted, which i don't think is correct. each group needs to follow the exact sequence of modes.
one idea is a helper like wait_for_status_change() where to pass the current mode, and the next mode as arguments. the status is then expected to stay in the current mode, until it changes to the next mode. if other modes is seen, it's an error

spec/support/command_helpers.rb
```diff
@@ -373,6 +373,14 @@ def wait_normal_control
     )
   end
 
+  def wait_normal_control_and_status
```
emiltin, line 376: maybe rename to something more destriptive like verify_startup_sequence?

spec/site/tlc/system_spec.rb
```diff
@@ -61,6 +61,37 @@
...
+    # 3. Wait for status response= stopped
+    # 4. Reconnect as Isolated_mode
+    # 5. Wait for status= ready
+    # 6. Send command to switch to normal controll
+    # 7. Wait for S0020 status "startup" 
+    # 8. Wait for S0001 status "eeeee"
+    # 9. Wait for S0001 status "ffffffff"
+    # 10. Wait for S0001 status "gggggg"
+    # 11. Wait for S0020 status "control" 
+    it 'M0004 restart and added status check', sxl: '>=1.0.7' do |example|
```
emiltin, line 75: confusing name. maybe something like 'goes through startup sequence after M0004 restart'

sweco-seguro:
I seem unable to push a new commit to this, am I still authorized to push commits?

emiltin:
> I seem unable to push a new commit to this, am I still authorized to push commits?

please confirm the invitation i send you last week

sweco-seguro:
Where can I find the invitation? I can't find any invitation in my mail, or here on github.

emiltin:
> Where can I find the invitation? I can't find any invitation in my mail, or here on github.

email. let my try to resend it.

emiltin:
i just re-invited you

sweco-seguro:
I seem to have changed the method names but missed the places where I actually call the method. Should be fixed now.

They haven't been run against any equipment at all as far as I know.

emiltin:
hi @mikaelbroms, I suggest adding a block argument to the rsmp gem method `subscribe_to_status` (and similar methods), that is called for each status item. this can be used to do custom validation of messages as they arrive,e.g. to check the sequence of signal groups.

The usage would be like this:
```ruby
message, result = @site.subscribe_to_status component, subscribe_list, options do |message, item|
  puts "group status: #{item["s"]}"
  true      # do custom check on the group sequences and return false if sequence is invalid
end
```

I can help with the rsmp gem code if you like.

emiltin:
closing in favour of #138, which includes updated versions of the additions in this PR

PR #119, opened 2021-09-22
Reuse proxies when reconnecting

### LINKED ISSUE #114: RSMP::NotReady when running equipment tests
When running tests against an EC-3,  a number of tests fails with RSMP::NotReady. This should be investigated.

```
 10) Traffic Light Controller Detector Logic forcing is set with M0008
      Failure/Error: @site.unsubscribe_to_status Validator.config['main_component'], unsubscribe_list
      
      RSMP::NotReady:
        RSMP::NotReady
      # ./vendor/bundle/ruby/3.0.0/gems/rsmp-0.2.1/lib/rsmp/site_proxy.rb:241:in `unsubscribe_to_status'
      # ./spec/support/status_helpers.rb:61:in `ensure in block in wait_for_status'
      # ./spec/support/status_helpers.rb:61:in `block in wait_for_status'
      # ./spec/support/log_helpers.rb:5:in `log_confirmation'
      # ./spec/support/status_helpers.rb:53:in `wait_for_status'
      # ./spec/support/command_helpers.rb:435:in `switch_detector_logic'
      # ./spec/site/tlc/detector_logics_spec.rb:47:in `block (5 levels) in <top (required)>'
      # ./spec/support/testee.rb:32:in `block in connected'
      # ./spec/support/testee.rb:101:in `block in within_reactor'
      # ./vendor/bundle/ruby/3.0.0/gems/async-1.29.1/lib/async/task.rb:260:in `block in make_fiber'
      # ------------------
      # --- Caused by: ---
      # Async::TimeoutError:
      #   execution expired
      #   ./vendor/bundle/ruby/3.0.0/gems/async-1.29.1/lib/async/task.rb:66:in `yield'
```

### LINKED ISSUE #144: Duplicate yellow flash tests
There seems to be duplicate test cases: "yellow flash can be activated with M0001" in spec/site/tlc/modes_spec.rb
* line 107
* line 162

emiltin:
This PR fixes #114.

The current implementation will create a SiteProxy when a site connects. When the connection is closed, the object is discarded. If the site then reconnects a new SiteProxy is created.

But a test might still hang on the the old SiteObject, and try to use to to send messages, which will cause NotReady erorrs. This typically happens when you wait for a message, but this times out, and you rescue the timeout and try to do some rsmp cleanup, like unsubscribing to statuses, reset the clock, etc.

The fix is is to keep SiteProxy objects when a site disconnects. When a site connects it will look for an existing SiteProxy with a matching RMSP site id, and reuse it if found. That way tests that hang on to the SiteProxy object can still use it once the site is reconnected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp'
+gem 'rsmp', path: '/Users/emiltin/Documents/code/rsmp'
 gem 'rspec'
 gem 'activesupport'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp', path: '/Users/emiltin/Documents/code/rsmp'
+gem 'rsmp', github: 'emiltin/rsmp'
 gem 'rspec'
 gem 'activesupport'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp', github: 'emiltin/rsmp'
+gem 'rsmp'
 gem 'rspec'
 gem 'activesupport'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,5 +1,6 @@
-PATH
-  remote: /Users/emiltin/Documents/code/rsmp
+GIT
+  remote: https://github.com/emiltin/rsmp.git
+  revision: f36855c1581ea69c569fadf3fe01ea99c5aa6ad8
   specs:
     rsmp (0.2.1)
       async (~> 1.29.1)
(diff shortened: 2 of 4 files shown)


emiltin:
For some reason this was problem was showing up mostly with the ITC-3 equipment.

emiltin:
if this pr works, a new version of the rsmp gem needs to be pushed to rubygems and the Gemfile updated

emiltin:
@otterdahl checking how a PR works with equipment is unfortunately a bit hard after we moved the equipment tests to rsmp_implementation. it seems it always uses the master branch from the rsmp_validator repo?

otterdahl:
> @otterdahl checking how a PR works with equipment is unfortunately a bit hard after we moved the equipment tests to rsmp_implementation. it seems it always uses the master branch from the rsmp_validator repo?

Right. I've added an issue here: https://github.com/rsmp-nordic/rsmp_implementations/issues/15

PR #133, opened 2021-11-03
try collection cancellation

emiltin:
testing PR against equipment, should trigger equipment tests

uses the collect_abort branch of the rsmp gem


emiltin:
merged

PR #141, opened 2022-01-20
Test clock

otterdahl:


spec/site/tlc/clock_spec.rb
```diff
@@ -49,6 +49,7 @@
       Validator::Site.connected do |task,supervisor,site|
         prepare task, site
         with_clock_set CLOCK do
+          sleep 10
```
emiltin, line 52: sleep() will block everything else, including responding to watchdog messages, timers, etc. instead use task.sleep()

otterdahl: Fixed. Thanks

otterdahl:
The conlusion is that the ITC-3 fails some clock tests because it doesn't change the clock immediately. It need some seconds to adjust. Adding a delay makes the test cases work, but I'm not sure if that is the best solution.

otterdahl:
We meed to verify the RSMP specification in this regard. Should CommandReponse be sent after the command has been executed or is it ok to send CommandResponse immediately?

Anyway. It might be a better approach to write the test cases for "clock" to use StatusSubscription for the clock and wait until the expected values has been reached. This is similar to how other test cases are written.

emiltin:
Thanks a lot for investigation this.

I think it's fully valid to use the CommandReponse, and very important that the expected behaviour of the CommandResponse is well defined. This kind of ambiguity is exactly what's causing incompatibilities.

I think modifying tests to make them pass by working around bugs in the equipment is counter-productive. Exposing issues is the purpose of our tool.

emiltin:
i suggest we include two tests:
one that verifies the clock change usign the command response
one that verifies the clock change using status subscription

emiltin:
closing, as the investigation is complete. i'll open a new issue with the proposed changes

PR #148, opened 2022-01-27
Verify that site goes back to normal control after M0001 timeout

mikaelbroms:
Test of new and changed test cases


spec/site/tlc/clock_spec.rb
```diff
@@ -157,7 +157,7 @@
       Validator::Site.connected do |task,supervisor,site|
         prepare task, site
         with_clock_set CLOCK do
-          result = set_functional_position 'NormalControl'
+          result = set_functional_position 'NormalControl' 0
```
emiltin, line 160: i think you're missing a comma here between 'NormalControl' and 0, to separate arguments?
the same several place below

spec/site/tlc/modes_spec.rb
```diff
@@ -166,6 +166,24 @@
...
     end
+
+    # Verify that site goes back to normal control after M0001 timeout
+    #
+    # 1. Given the site is connected
+    # 2. Send the control command to switch to Yellow flash
+    # 3. Wait for status Yellow flash
+    # 4. Send command to switch to normal control
+    # 5. Wait for status "Yellow flash" = false, "Controller starting"= false, "Controller on"= true"
+    specify 'yellow flash can be activated with M0001', sxl: '>=1.0.7' do |example|
```
emiltin, line 177: the spec name must be changed, if this is a new test. another test already exist with this name. it should be describtive and work as a continuation of the phrase 'Traffic Light Controller Operational ...'

spec/site/tlc/modes_spec.rb
```diff
@@ -166,6 +166,24 @@
...
+      Validator::Site.connected do |task,supervisor,site|
+        prepare task, site
+        # timeout = Validator.config['timeouts']['m0001']
+        timeout = 1
+        switch_yellow_flash timeout
+        task.sleep(120)
```
emiltin, line 183: that's the purpose of this long delay? generally we should avoid delays, because they makes running tests very slow.
in case we really need a delay, it must be configurable, ie. fetched from the YAML config file.

spec/site/tlc/signal_groups_spec.rb
```diff
@@ -101,10 +101,10 @@
       Validator::Site.connected do |task,supervisor,site|
         prepare task, site
         verify_startup_sequence do
-          switch_yellow_flash
+          switch_yellow_flash 0
```
emiltin, line 104: what does 0 signify here?

emiltin:
github actions workflow files are located in .github/workflows/

In the workflow, you can see the command to riun tests, eg:
https://github.com/rsmp-nordic/rsmp_validator/blob/master/.github/workflows/swarco_itc3.yml#L31

You can modify that to run specific tests.

Just be sure to revert chagnes to workflow files before merging.

If you just want to run tests against the ruby tlc emulator, it's much easier to just do it locally.

emiltin:
you can disable workflows by e.g. adding an 'if':

```yaml
jobs:
  test:
    if: ${{ false }}      # disable for now
```

.github/workflows/gem_supervisor.yml
```diff
@@ -15,37 +15,4 @@ on:
...
-      with:
-        name: rspec validation
-        path: log/validation_ruby_*.log
-    - name: Upload simulator.log
-      if: always()
-      uses: actions/upload-artifact@v2
-      with:
-        name: simulator log
-        path: simulator.log
+    if: ${{ false }}      # disable for now
```
emiltin, line 18: you can leave the rest of the code, you don't have to delete it.

emiltin:
I strongly suggest running the test locally against the TLC emulator before pushing here. it's much faster to catch and fix things like syntax errors that way.

spec/support/command_helpers.rb
```diff
@@ -83,13 +83,13 @@ def set_traffic_situation ts
   end
 
   # Set functional position
-  def set_functional_position status
+  def set_functional_position status timeout
```
emiltin, line 86: you have a syntax error here. you need comma between arguments

PR #149, opened 2022-01-28
refactor async usage, update rsmp and related specs

emiltin:
Refactor use of Async, update rsmp to branch that also updates Async usage.
All specs now pass with the TLC emulator, including tests that restart the emulator.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'rsmp'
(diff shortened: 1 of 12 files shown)


emiltin:
Refactor seems pretty solid.
The Ruby TLC and Supervisor passes all tests. EC-2, ITC-2 and ITC-3 fails a bunch of tests, as usual.

emiltin:
trying another run

PR #153, opened 2022-02-02
test all configured detector logics, not just the first

emiltin:
a PR makes it easier to follow the work in the branch

### DIFF
--- a/spec/support/command_helpers.rb
+++ b/spec/support/command_helpers.rb
@@ -506,22 +506,22 @@ module Validator::CommandHelpers
 
   def switch_detector_logic
     indx = 0
-    component = Validator.config['components']['detector_logic'].keys[indx]
-
-    force_detector_logic component, mode:'True'
-    Validator.config['main_component'] = Validator.config['main_component']
-    wait_for_status(@task,
-      "activate detector logic #{component}",
-      [{'sCI'=>'S0002','n'=>'detectorlogicstatus','s'=>/^.{#{indx}}1/}]
-    )
-    
-
-    force_detector_logic component, mode:'False'
-    Validator.config['main_component'] = Validator.config['main_component']
-    wait_for_status(@task,
-      "deactivate detector logic #{component}",
-      [{'sCI'=>'S0002','n'=>'detectorlogicstatus','s'=>/^.{#{indx}}0/}]
-    )
+    Validator.config['components']['detector_logic'].keys.each do |component|
+      force_detector_logic component, mode:'True'
+      Validator.config['main_component'] = Validator.config['main_component']
+      wait_for_status(@task,
+        "activate detector logic #{component}",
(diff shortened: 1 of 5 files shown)


spec/support/command_helpers.rb
```diff
@@ -502,22 +502,22 @@ def switch_input indx
...
+        [{'sCI'=>'S0002','n'=>'detectorlogicstatus','s'=>/^.{#{indx}}1/}]
+      )
+      
+      force_detector_logic component, mode:'False'
+      Validator.config['main_component'] = Validator.config['main_component']
+      wait_for_status(@task,
+        "deactivate detector logic #{component}",
+        [{'sCI'=>'S0002','n'=>'detectorlogicstatus','s'=>/^.{#{indx}}0/}]
+      )
+      indx = indx + 1
```
emiltin, line 519: you can use each_with_index, then you don't have to manually count

spec/support/command_helpers.rb
```diff
@@ -502,22 +502,22 @@ def switch_input indx
...
-    )
-    
-
-    force_detector_logic component, mode:'False'
-    Validator.config['main_component'] = Validator.config['main_component']
-    wait_for_status(@task,
-      "deactivate detector logic #{component}",
-      [{'sCI'=>'S0002','n'=>'detectorlogicstatus','s'=>/^.{#{indx}}0/}]
-    )
+    Validator.config['components']['detector_logic'].keys.each do |component|
```
emiltin, line 505: use each_with_index

emiltin:
looks good to me.
are there other helper methods that uses detector logics, that should be updated

spec/support/command_helpers.rb
otterdahl: Should be fixed now

mikaelbroms:
'forcing is set with M0019' works with 2 pcs detector logics KK + AG9998 = 001DL001: KK + AG9998 = 001DL002:

if only force_detector_logic component, mode: 'True' runs in
def switch_detector_logic

if also force_detector_logic component, mode: 'False' runs in
def switch_detector_logic

then it does not work with one or more detector logics

mikaelbroms:
It seems that 'force_detector_logic component, mode:'False' not work in TLC.
Or that the subscription
wait_for_status
 [{'sCI'=>'S0002','n'=>'detectorlogicstatus','s'=>/^.{#{indx}}0/}]
not goes to zero within the timeout

otterdahl:
I can confirm that is doesn't work. Might be because of some configuration setting in the TLC

emiltin:
Is there more to do here? It seems you have confirmed that the problem is in th TLC. In that case I think we can delete this PR without merging?

mikaelbroms:
The purpose for this PR is to test multi detector logics. The code for testning multi detektor logics is working. 

Skickat från min iPhone

otterdahl:
I think this PR works. I'm merging this

PR #154, opened 2022-02-03
M0019 unforce

emiltin:


otterdahl:
> Maybe the regex /^.{#{input}}1/} is wrong

Yes, it is incorrect. It actually checks input 102 and not input 101 as intended.

spec/site/tlc/io_spec.rb
```diff
@@ -21,11 +21,51 @@
       # 3. Wait for status = true  
       specify 'forcing is set with M0019', sxl: '>=1.0.13' do |example|
         Validator::Site.connected do |task,supervisor,site|
-          status = 'False'
-          input = 1
-          inputValue = 'True'
+
           prepare task, site
+
+          input = 101
```
otterdahl, line 27: Consider reading input from config file `Validator.config['items']['inputs']` instead of using a hard coded value

For inspiration, check out the test case "is activated with M0006":

```ruby
inputs = Validator.config['items']['inputs']
skip("No inputs configured") if inputs.nil? || inputs.empty?
    prepare task, site
    inputs.each { |input| switch_input input }
end
```

otterdahl:
We're hitting this issue again: https://github.com/rsmp-nordic/rsmp/issues/35

Issue #142, opened 2022-01-21
M0001 Add test case to verify that "timeout" works as intended


otterdahl:
Currently the M0001 is only testing that yellow flash and dark mode works.

There should also be test cases (1 or 2?) that checks that the "timeout" property works. If this is set to >0 then the TLC should revert to previous function position (always NormalMode?) when this time is reached.

otterdahl:
A test case has been merged. Closing this

PR #155, opened 2022-02-03
M0001 verify timeout

emiltin:
Test 'timeout' property of M0001 #142 

Currently the M0001 is only testing that yellow flash and dark mode works. There should also be test cases (1 or 2?) that checks that the "timeout" property works. If this is set to >0 then the TLC should revert to previous function position (always NormalMode?) when this time is reached.

### DIFF
--- a/spec/site/tlc/modes_spec.rb
+++ b/spec/site/tlc/modes_spec.rb
@@ -179,7 +179,7 @@ RSpec.describe 'Site::Traffic Light Controller' do
     Validator::Site.connected do |task,supervisor,site|
       prepare task, site
       switch_yellow_flash 1
-      task.sleep(70)
+      task.sleep(90)
       wait_normal_control
     end
   end  
--- a/spec/site/tlc/modes_spec.rb
+++ b/spec/site/tlc/modes_spec.rb
@@ -171,13 +171,14 @@ RSpec.describe 'Site::Traffic Light Controller' do
   # Verify that we can activate yellow flash and after 1 minute goes back to NormalControl
   #
   # 1. Given the site is connected
+  # 2. Send the control command to switch to Normal Control, and wait for this
   # 2. Send the control command to switch to Yellow flash
   # 3. Wait for status Yellow flash
-  # 4. Send command to switch to normal control
-  # 5. Wait for status "Yellow flash" = false, "Controller starting"= false, "Controller on"= true"
+  # 5. Wait for automatic revert to Normal Control
   specify 'yellow flash can be activated with M0001 and goes back to NormalControl after one minute', sxl: '>=1.0.7' do |example|
     Validator::Site.connected do |task,supervisor,site|
       prepare task, site
+      switch_normal_control
       switch_yellow_flash 1
       wait_normal_control 120
     end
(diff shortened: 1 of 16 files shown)


emiltin:
the gem tlc test fail. it's because the emulator doesn't yet implment the yellow flash timeout - so the test seems to work correctly.

mikaelbroms:
yellow flash can be activated with M0001 and goes back to NormalControl after one minute
I have removed task.sleep(90)
added optional parameter 'timeot' to helper method wait_normal_control 
pass 120 seconds as timeout to method wait_normal_control
The test doesn´t pass.

Error in detailed log: undefined local variable or method `e'
https://github.com/rsmp-nordic/rsmp_validator/runs/5230295998?check_suite_focus=true

otterdahl:
> Error in detailed log: undefined local variable or method `e' https://github.com/rsmp-nordic/rsmp_validator/runs/5230295998?check_suite_focus=true

I've opened an issue at: https://github.com/rsmp-nordic/rsmp/issues/35

emiltin:
> the gem tlc test fail. it's because the emulator doesn't yet implment the yellow flash timeout - so the test seems to work correctly.

the gem tlc now supports yellow flash timeout, and the test pass

emiltin:
rebased on master

spec/site/tlc/modes_spec.rb
```diff
@@ -179,8 +179,7 @@
     Validator::Site.connected do |task,supervisor,site|
       prepare task, site
       switch_yellow_flash 1
-      task.sleep(90)
-      wait_normal_control
+      wait_normal_control 120
```
emiltin, line 182: the yellow flash timeout used in 1 minute, ie. 60 seconds. i guess a little bit of buffer is reasonable, but do we need 30 seconds? i think this extra time should be a test config, not hardcoded

spec/site/tlc/modes_spec.rb
```diff
@@ -107,7 +107,7 @@
     specify 'yellow flash can be activated with M0001', sxl: '>=1.0.7' do |example|
       Validator::Site.connected do |task,supervisor,site|
         prepare task, site
-        switch_yellow_flash
+        switch_yellow_flash 0
```
emiltin, line 110: i would prefer not to ad all these '0' arguments. let's use a defualt argument to avoid them

emiltin: i fixed this

emiltin:
the gem tlc now supports m0001 timeout and passes all tests

emiltin:
i added a switch to normal mode at the start of the test. otherwise the test fails if the tlc is not in normal control when the test is run, because the timeout will switch back to something else than normal control.

emiltin:
> Error in detailed log: undefined local variable or method `e'
> https://github.com/rsmp-nordic/rsmp_validator/runs/5230295998?check_suite_focus=true

this was also fixed

otterdahl:
- ITC-3: Fail due to #168
- ITC-2: It works!
- EC-2: Unable to verify

I'm aiming to clean up this PR and merge it.

PR #162, opened 2022-02-18
Investigate why M0001 test cases fail on ITC2

otterdahl:


emiltin:
Is there more to do here? It seems you have confirmed that the problem is in th TLC. In that case I think we can delete this PR without merging?

otterdahl:
> Is there any conclusion in this investigation?

Not done yet. M0001 used to work at some point. Unclear why it doesn't work any longer.

otterdahl:
Using intersection=1 causes Dark mode to fail. At least using the current ITC-2 configuration.
NormalControl still fails for some unknown reason

otterdahl:
M0001 now works. 'timeout' needed to be increased to 180 s

emiltin:
merge?

otterdahl:
> merge?

Yes. The timeout was adjusted for ITC-3. This change has been merged.

PR #165, opened 2022-02-23
try setup-ruby for the itc2 workflow

emiltin:


emiltin:
Fails with:
```
Modifying PATH
  Entries added to PATH to use selected Ruby:
    /opt/hostedtoolcache/Ruby/3.1.0/x64/bin
Error: ENOENT: no such file or directory, mkdir
```

There might be some tips at:
https://github.com/ruby/setup-ruby#using-self-hosted-runners


One reason I would like to use ruby-install is that the version of ruby to use is read from .ruby-version.
And it's nice be install ruby and gems in the same way for all workflows.

emiltin:
putting this on hold. would be nice to have a consistent setup, but we have other things to do at the moment

PR #167, opened 2022-03-02
Investigate why M0001 test cases fail on ITC3

otterdahl:
The M0001 works when testing directly with the RSMP simulator.
Why does it fail when using RSpec?

### DIFF
--- a/.github/workflows/dynniq_ec2.yml
+++ b/.github/workflows/dynniq_ec2.yml
@@ -15,6 +15,8 @@ on:
 
 jobs:
   test:
+    if: ${{ false }}
+
     runs-on: [ self-hosted, Windows, X64, ec2 ]
     steps:
     - name: Check out repository
--- a/.github/workflows/swarco_itc3.yml
+++ b/.github/workflows/swarco_itc3.yml
@@ -28,7 +28,7 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems
 
     - name: Run tests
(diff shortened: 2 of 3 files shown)


otterdahl:
It fails because there is a check for S0007 (Switched on) that returns "False" during YellowFlash => NormalMode,
It is because the ITC-3 is running in 'off' mode, e.g. no output to the lamps, since none are connected. When trying to turning the controller 'On' the TLC reports lots of errors. No lamps are working.

otterdahl:
Either lamps needs to be connected to the TLC, or the alarms needs to suppressed using an updated configuration of the TLC.

otterdahl:
I've opened a separate issue at https://github.com/rsmp-nordic/rsmp_validator/issues/168. Closing this

otterdahl:
> It fails because there is a check for S0007 (Switched on) that returns "False" during YellowFlash => NormalMode, It is because the ITC-3 is running in 'off' mode, e.g. no output to the lamps, since none are connected. When trying to turning the controller 'On' the TLC reports lots of errors. No lamps are working.

This is cirumvented by setting the TLC into simulation mode

otterdahl:
The M0001 contain a an argument "timeout", but rsmp_validator doesn't seem to send it

emiltin:
timeout=0 is now used when sending M0001. are tests expected to pass?

emiltin:
fixed in #183

PR #169, opened 2022-03-07
disable emergency routes using index

mikaelbroms:


otterdahl:
Cannot verify with ITC-2/ITC-3 due to schema error during subscription to S0006.

```
 sCI":"S0006","n":"emergencystage","s":"",
```
emergencystage cannot be an empty string

emiltin:
> Cannot verify with ITC-2/ITC-3 due to schema error during subscription to S0006.
> 
> ```
>  sCI":"S0006","n":"emergencystage","s":"",
> ```
> 
> emergencystage cannot be an empty string

is the 'status' attribute not send?

looking at the spec at:
https://rsmp-nordic.org/rsmp_specifications/rsmp_sxl_traffic_lights/1.0.15/sxl_traffic_light_controller.html#s0006

it does not state whether emergencystage must be present if status is False.

emiltin:
can be merged - it's correct that tests fail

emiltin:
rebased on master, let's see how the checks fare

PR #170, opened 2022-03-08
add new test case for alarm test

mikaelbroms:


spec/site/tlc/alarm_spec.rb
```diff
@@ -20,7 +20,64 @@
...
+    # 2. We trigger an alarm by setting S0003 input 152 to True
+    # 3. Then we should receive an active alarm
+    # 4. We remove an alarm by setting S0003 input 152 to False
+    # 5. Then we should receive an inactive alarm
+
+    specify 'A0302 is raised when a detector logic is faulty', :script, sxl: '>=1.0.7' do |example|
+      Validator::Site.connected do |task,supervisor,site|
+
+        # Alarm is raised by setting S0003 input 152 to True
+        ensure_input 'True', 152 do
```
emiltin, line 35: using ensure_input with with a block here is a bit odd, since the function does not take a block as an argument

spec/site/tlc/alarm_spec.rb
```diff
@@ -28,15 +28,15 @@
     # 4. We remove an alarm by setting S0003 input 152 to False
     # 5. Then we should receive an inactive alarm
 
-    specify 'A0302 is raised when a detector logic is faulty', :script, sxl: '>=1.0.7' do |example|
+    specify 'A0301 is raised when S0003 input 152 is forced to True', :script, sxl: '>=1.0.7' do |example|
```
emiltin, line 31: the hard coded number 152 should be read from the test config

spec/site/tlc/alarm_spec.rb
```diff
@@ -20,7 +20,66 @@
...
+    # 3. Then we should receive an active alarm
+    # 4. We remove an alarm by setting S0003 input 152 to False
+    # 5. Then we should receive an inactive alarm
+
+    specify 'A0301 is raised when S0003 input 152 is forced to True', :script, sxl: '>=1.0.7' do |example|
+      Validator::Site.connected do |task,supervisor,site|
+
+        # Alarm is raised by setting S0003 input 152 to True
+        prepare task, site
+        ensure_input('True', 152) do
```
emiltin, line 36: as mentioned before, using block with ensure_input() doesn't make sense, since the method doesn't use the block

spec/site/tlc/alarm_spec.rb
```diff
@@ -20,7 +20,66 @@
...
+          site.log "Waiting for alarm", level: :test
+          start_time = Time.now
+          alarm_code_id = 'A0301'
+          collector = site.collect_alarms num: 1, component: component, aCId: alarm_code_id,
```
emiltin, line 40: i suspect collect_alarms() is not working. alarm tests have been disabled for a while, and i did a major refactoring of the Collector classes. i think we need a proper AlarmCollector class. i'll look at that

emiltin: did you try to run the test? does it collect the expected method?

mikaelbroms: Perfect! I run the test. see https://github.com/rsmp-nordic/rsmp_validator/runs/5479066165?check_suite_focus=truehttps://github.com/rsmp-nordic/rsmp_validator/runs/5479066165?check_suite_focus=true

emiltin: well, it seems to pass, but the detailed log for that test shows no Alarm being received:
https://github.com/rsmp-nordic/rsmp_validator/runs/5479066165?check_suite_focus=truehttps://github.com/rsmp-nordic/rsmp_validator/runs/5479066165?check_suite_focus=true#step:5:147

if you run this against the local ruby TLC emulator, it should fail.

i think what's happening is this: most of the block is passed as blocks to ensure_input(), but that method doesn't use the block for anything and doesn't call it. so most of the code is never run.

emiltin:
hi @mikaelbroms, see the PR #171. it includes fixes the the new alarm test. you can either merge it (to you new_alarm_test branch), or use it as inspiration.

once the code was activate, there was a number of smaller issues that had to be fixed

emiltin:
hi @mikaelbroms there are several other things that needs to be fixed, see my pr #171. 
I suggest you try to run the test locally first against the ruby tcl emulator. since the ruby emulator can't (yet) raise an alarm when an input is activated, you should see the test fail because an alarm was never received.

emiltin:
pr #171 merged.
i would like to add ability to raise alarms in the tlc emulator, so it passes the test

emiltin:
the test shows that the itc-3 does not send any alarm after input 152 is forced to True.

emiltin:
forcing the input, rather than just setting, means that the TLC now raises an alarm.

the check for the rvs values fails because we expect:

```ruby
        expected_rvs = [
          {"n"=>"detector","v"=>"1"},
          {"n"=>"type","v"=>"loop"},
          {"n"=>"errormode","v"=>"on"},
          {"n"=>"manual","v"=>"True"},
          {"n"=>"logicerror","v"=>"always_off"}
        ]
```

but the alarm has:

detector=3
errormode=off
manual = False


another thing i noticed is that the same alarm is received several times, with no changes, which causes the validator to respond with NotAck.

emiltin:
I see a consistent disconnect happening every time input 152 is forced to True, which seems like a bug to me?

https://github.com/rsmp-nordic/rsmp_validator/runs/5518047624?check_suite_focus=true#step:5:107

emiltin:
The alarm collector expected  'aSp' =>  'Issue' but the ITC-3 uses the incorrect value 'issue' with a lowercase i.
We didn't notice that because the schema validation is currently turned off for alarms on the ITC-3, exactly because of this issue.
I removed the check for aSp for now.

emiltin:
the same issue is present when matching aS with 'Active', the ITC-3 uses the incorrect 'issue'. But since that's part of the test, we need to use case-insensitive matching, or simply conclude that the ITC-3 fails this test.

emiltin:
The ITC-3 uses incorrect casing for most of the alarm attributes. I think the correct way to deal with this is to first of all rely on the json schema validation. That will catch all problems with casing as they arrive.
To also test the alarm behaviour, case insensitive reges is not used in the alarm test. As long as the json scheme validation is disabled for alarms (as is currently the case), the test passes.

emiltin:
merging. we still need to update the ITC-2 config and check how the test works, but we can do that in a separate issue

PR #174, opened 2022-03-28
Alarm acknowledgement

mikaelbroms:


config/gem_tlc.yaml
```diff
@@ -8,7 +8,7 @@ intervals:
 timeouts:
   watchdog: 1
   acknowledgement: 1
-  connect: 1
+  connect: 180
```
emiltin, line 11: remember to set this back to 1

spec/site/tlc/alarm_spec.rb
```diff
@@ -68,13 +68,44 @@
             alarm_time = Time.parse(alarm.attributes["aTs"])
             expect(alarm_time).to be_within(1.minute).of Time.now.utc
             log "Alarm #{alarm_code_id} is now #{alarm_status.inspect}"
+
+            if (alarm_status.inspect == 'True')
```
emiltin, line 72: this will never we true. alarm_status is the value, not the key, of the key-value pairs. i think you want:

```ruby
if input_value == 'True'
```

emiltin:
spec_helper.rb in the master branch now requires both active_support and active_support/time. it seems that's required.
i rebased this branch on master, and squashed the commits.

mikaelbroms:
The TLC responds with "aSp":"issue" when it should respond with "aSp":"acknowledge" according to the specification.

mikaelbroms:
The test case works if  "aSp is set to :"issue", but this is incorrect.

config/gem_tlc.yaml
mikaelbroms: done

mikaelbroms:
@otterdahl and I found that the test case work with ITC-3, but the TLC has issue with "aSp:Issue". But otherwise the test case seems to work.

spec/site/tlc/alarm_spec.rb
```diff
@@ -40,6 +40,64 @@
...
+
+        # the rsmp spec requires a specific casing of enums, but some equipment uses incorrect casing.
+        # incorrect casing is usualle cuaght by the json schema validation, but in case this is disabled,
+        # we use case-insensitive regex patterns so that the tests can still be run
+        mapping = {
+          'True' => /Active/i,    # when input is forced to True, the alarm should become active
+          'False' => /inActive/i  # when input is forced to False, the alarm should become inactive
+        }
+
+        mapping.each_pair do |input_value, alarm_status|
```
emiltin, line 58: i think the test can be simplified a bit, since we're not testing the activation? so instead of looping over input=>alarm pairs, just do:

- force input to false
- force input to true
- wait for an alarm
- acknowledge the alarm
- check that the alarm status becomes acknowledged
- release input

emiltin:
stale. going to open a new PR for alarm acknowledgement

PR #173, opened 2022-03-24
Alarm suspend

emiltin:
shows how to send an alarm suspend message

PR #176, opened 2022-03-30
Suspend alarm

mikaelbroms:


emiltin:
did you see the PR #173 "alarm suspend"?

spec/site/tlc/alarm_spec.rb
```diff
@@ -74,6 +74,40 @@
         end
       end
     end
+    
+    it 'can be suspended', :script, sxl: '>=1.0.7' do |example|
+      
+      alarm_code_id = 'A0301'   # what alarm to expect
+      component = 'KK+AG9998=001DL003'
```
emiltin, line 88: component ids should be read from the config, not hardcoded into tests. they differ depending on the equipment you test. the detector logic component ids are already in the config, but in case you need a specific DL for this test, you can the detector logic nr as a specific config element.

spec/site/tlc/alarm_spec.rb
```diff
@@ -74,6 +74,40 @@
...
+      alarm_status = /inActive/i
+      timeout  = Validator.config['timeouts']['alarm']
+
+      Validator::Site.connected do |task,supervisor,site|
+        log "Check that alarm suspend #{alarm_code_id} can be send and confirmed"
+        m_id = RSMP::Message.make_m_id  # generate a message id, that can be used to listen for repsonses
+        alarm = RSMP::AlarmSuspend.new(
```
emiltin, line 96: is is valid to suspend an alarm that is not active? i guess it might be, but worth checking.
a potential problem here is that if the alarm is already suspended, you will not get an alarm when suspending it. to make it more robust, i suggest doing an initial resume before moving on the the actual test.

emiltin: another potential issue is that you expect the alarm status to be /inActive/i, but that will fail if the alarm is active when the test is running.

emiltin:
hi @mikaelbroms,
i updated the rsmp gem, so that it responds to alarm suspend/resume. it makes it easier to see what's going on with the tcl emulator. i also updated the test to fix a few issues.

emiltin:
hi @mikaelbroms, @otterdahl.
I updated to rsmp_schema, so that it correctly validates Alarm suspend/resume/suspended/resumed message. the rsmp gem now also maintain alarm states on both the site and supervisor site.

I updated the alarm suspend test to benefit from this, and it now passes with the ruby tlc emulator.

but the ITC-3 and the ITC-2 fails the test, because it sends alarm messages with enums that use different casing that what's  required by the spec.

it's debatable whether this this incorrect, since case sensitivity is has now been stated explicitely in the spec so far, but will be in the upcoming version of core.

so we can either:
a) update the rsmp core schema up to 3.1.5 to allow variations of casing. for example, 'inActive', 'inactive' Inactive', InActive', or perhaps any casing, although that's hard to with with json schema.
b) conclude that the equipment fails this test. from the upcoming core 3.2, there will be no doubt that they must use the correct casing for all enums.

btw, perhaps we should strengthen the fomulation in the upcoming core, by mentioning that all enum values must use the exact casing defined in the spec? i thin k we currently only states that parsing must be done case sensitive.

emiltin:
@mikaelbroms @otterdahl is this PR ready to me merge, or would you like to add/change things?

emiltin:
superseded by #173

PR #183, opened 2022-05-19
use 0 as default for timeout when switcing to yellow flash

emiltin:
fixes missing timout=1 in M0001 commands


emiltin:
looks good to me

PR #185, opened 2022-06-14
Increase startup sequence timeout for ITC-3

otterdahl:
Investigate if increase timeout for startup sequence makes ITC-3 pass the test "follow startup sequence after yellow flash"

### DIFF
--- a/.github/workflows/swarco_itc3.yml
+++ b/.github/workflows/swarco_itc3.yml
@@ -28,7 +28,7 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems
 
     - name: Run tests
-      run: bundle exec rspec --format Validator::Brief --format Validator::Details --out log/validation.log spec/site/tlc/signal_groups_spec.rb:100
+      run: bundle exec rspec --format Validator::Brief --format Validator::Details --out log/validation.log spec/site
       env:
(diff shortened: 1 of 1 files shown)


emiltin:
the expected startup sequence is defined in config/swarco_itc3.yaml:

```yaml
startup_sequence: 'efg'
```

is 'efg' correct? it looks like all groups switch to 'e' and never progresses.

otterdahl:
> is 'efg' correct? it looks like all groups switch to 'e' and never progresses.

Yes. I think it is. The controller gets stuck in dark mode.

<html>
<body>


e | Start-up interval 1 | Dark or yellow flash |  
-- | -- | -- | --
f | Start-up interval 2 | Vehicles: yellow, Pedestrians: red |  
g | Start-up interval 3 | Red |  


</body>
</html>

I've looked at the web interface of the controller and it seems to have a 'major error' causing this

emiltin:
> We need to switch to normal control without waiting for startup mode to be over

sounds plausiable, we should try to change the test a bit.

otterdahl:
> > We need to switch to normal control without waiting for startup mode to be over
> 
> sounds plausiable, we should try to change the test a bit.

Yes, it fixes it.

emiltin:
nice!

PR #190, opened 2022-09-02
signal priority tests M0022 and S0033

emiltin:
Add tests for M0022 (requesting priority) and S0033 (priority status).

S0033 is the first RSMP status message that uses the native JSON array type to return data.
The test for M0022 does the following:

1. Given the site is connected
2. And we subscribe to signal priority status updates
3. When we send a signal priority request
4. Then we should receive status updates
5. And the request should go through the correct states

The priority request is expected to go through the following states:

received > activated > completed

The tests only run on equipment configured as using sxl >= 1.1. 

I consider this a good starting point. We can consider other ways we might want to test the priority messages.


emiltin:
As usual, the ruby gem has been updated to pass all tests. It implements a simplistic behaviour for M0022:

- when an M0022 priority request is received, it creates a new request int he 'received' state.
- 0.5 second later, the request moves to the state 'activated'
- 0.5 second later, the request moves to the state 'completed'
- 1.5 seconds later, the request is removed

S0033 returns the list of priority requests as a native JSON array, as required by the 1.1 spec.

otterdahl:
What is the status of this PR? Is it ready for merge?

PR #196, opened 2022-09-23
Version handling

emiltin:


emiltin:
when testing the gem tlc emulator, don't restrict testing to a specific sxl version - just use the latest by default

PR #201, opened 2022-09-30
Supervisor connect failures

emiltin:
investigating supervisor test failing on windows


emiltin:
The async-io gem is not fully compatible with Windows. Our support for Windows is therefore experimental only.

But you can install WSL (Windows Subsystem for Linux) on Windows, and then install and run the linux version of the validator inside WSL.

https://learn.microsoft.com/en-us/windows/wsl/install

Perhaps we should add a Windows machine with WSL as a test target.

PR #228, opened 2023-03-31
Fix 219

emiltin:
fixes #219

PR #231, opened 2023-04-11
Add support for 'source' field in several statuses

otterdahl:
In SXL 1.1 the 'source' field was added to several statuses, S0007, S0008, S0009, S0010, S0011, S0012, S0014, S0015, S0032.

This PR adds separate tests cases for those statuses if >= SXL 1.1

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -32,7 +32,7 @@ GEM
     nio4r (2.5.9)
     rake (13.0.6)
     regexp_parser (2.7.0)
-    rsmp (0.17.1)
+    rsmp (0.17.2)
       async (~> 1.30.3)
       async-io (~> 1.33.0)
       colorize (~> 0.8.1)
--- a/spec/site/tlc/modes_spec.rb
+++ b/spec/site/tlc/modes_spec.rb
@@ -47,9 +47,9 @@ RSpec.describe 'Site::Traffic Light Controller' do
     specify 'switched on is read with S0007', sxl: '>=1.0.7' do |example|
       Validator::Site.connected do |task,supervisor,site|
         if RSMP::Proxy.version_meets_requirement?( site.sxl_version, '>=1.1' )
-          status_list = { S0007: [:status,:intersection] }
-        else
           status_list = { S0007: [:status,:intersection,:source] }
+        else
+          status_list = { S0007: [:status,:intersection] }
         end
         request_status_and_confirm site, "controller switch on (dark mode=off)", status_list
       end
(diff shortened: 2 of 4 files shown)


emiltin:
See also my fix for sOc use in #230, it's very similare - use sOc attribute or not, depending on the negotiated version (core in this case, where you would use the sxl version).

otterdahl:
Thank you for the feedback

otterdahl:
The gem TLC doesn't seem to support 'source' yet.

otterdahl:
Done. What is left to do is to adapt according to https://github.com/rsmp-nordic/rsmp_validator/pull/232. Merge into single test cases and use `RSMP::Proxy.version_requirement_met?`

otterdahl:
It still fails. I wonder why...

emiltin:
try this update (note: force push)

emiltin:
gem tlc tests passes. looks good to you?

otterdahl:
Yes. I looks good. I'm just a little confused why `RSMP::Proxy.version_meets_requirement?( site.sxl_version, '>=1.1' )` returns false when SXL 1.1...

emiltin:
Is this logic correct? Source is only included if sxl is less than 1.1. Wasn't source added in 1.1?

```ruby
        if RSMP::Proxy.version_meets_requirement?( site.sxl_version, '>=1.1' )
          status_list = { S0007: [:status,:intersection] }
        else
          status_list = { S0007: [:status,:intersection,:source] }
        end
```

emiltin:
I tihnk there's an issue because the gem tlc does not handle the source attribute for e.g. S0007. The source is set to an empty string, schema validation fails when sending, and the validator test fails because no response is received. If i. hange the gem tlc so set source='other' the test passes.

otterdahl:
> Is this logic correct? Source is only included if sxl is less than 1.1. Wasn't source added in 1.1?

Yes, it was added in 1.1. I was confused why it tests suddenly succeed but I get it now.

emiltin:
the logic was inverted in a few places, should be fixed now. i also updated the gem to correct handle the source attribute in all status responses. it uses 'startup' initially, and chances to 'forced' if set through a command.

btw, if you switch functional position with a timeout, when it then reverts, what source is that? i used calender_clock

otterdahl:
Yes, calendar_clock is fine

emiltin:
Guess itc-2 and itc-3 are dead/offline? It would have been nice to see how that handle this update of the validator tests. Anything else before we merge?

otterdahl:
Yeah, the ITC-3 and EC-2 are till offline for the same reason as previously [discussed here](https://github.com/rsmp-nordic/rsmp_validator/pull/228).

I think we're all ok to merge

PR #232, opened 2023-04-11
Add test cases for the updated S0091 & S0092 in SXL 1.1

otterdahl:
In SXL 1.1 the S0091 & and S0092 were updated. It now only have 'user' attribute and not both 'user and 'status'.


emiltin:
The reason is that if you don't use the restrict_testing option, all tests are run, and then one of your tests fail, depending on what sxl version the gem tlc is running.

Perhaps we shoud rename 'restrict_testing' to something like 'versions' and make it a required option.

emiltin:
I suggest having just one test case, instead one for <1.1 and one for >=1.1, since it's really the same test and mechanics:

```ruby
    specify 'operator logged in/out of OP-panel is read with S0091', sxl: '>=1.0.7' do |example|
        if RSMP::Proxy.version_requirement_met?( '>=1.1', site.sxl_version )
          status_list = { S0091: [:user, :status] }
        else
          status_list = { S0091: [:user] }
        end
        request_status_and_confirm site, "operator logged in/out OP-panel", status_list
      end
    end
```

However, we cannot read site.sxl_version with access to the site object. I suggest we change the request_status_and_confirm() helper so it doesn't include Validator::Site.connected, as shown. above. I also think it's nice to have Validator::Site.connected in each test case, to more clearly show what's going on in the test.

otterdahl:
> The reason is that if you don't use the restrict_testing option, all tests are run, and then one of your tests fail, depending on what sxl version the gem tlc is running.

Is there any way to configure which sxl version the gem tlc is running or it always the latest version?

otterdahl:
> I suggest having just one test case, instead one for <1.1 and one for >=1.1, since it's really the same test and mechanics:

Sounds good.

> We cannot read site.sxl_version with access to the site object. I suggest we change the request_status_and_confirm() helper so it doesn't include Validator::Site.connected, as shown. above. I also think it's nice to have Validator::Site.connected in each test case, to more clearly show what's going on in the test.

OK. Should be add a separate pull request for the new version of request_status_and_confirm() before finishing this PR?

emiltin:
> > The reason is that if you don't use the restrict_testing option, all tests are run, and then one of your tests fail, depending on what sxl version the gem tlc is running.
> 
> Is there any way to configure which sxl version the gem tlc is running or it always the latest version?

You can configure it, otherwise it will use the latest. The validator read the gem tlc config from config/simulator/tlc.yaml and as you can see it does not include core or sxl version at the moment. For an example of a config that sets it, see https://github.com/rsmp-nordic/rsmp/blob/master/config/tlc.yaml. Perhaps we should be explicit in out validator gem tlc config.

emiltin:
> > I suggest having just one test case, instead one for <1.1 and one for >=1.1, since it's really the same test and mechanics:
> 
> Sounds good.
> 
> > We cannot read site.sxl_version with access to the site object. I suggest we change the request_status_and_confirm() helper so it doesn't include Validator::Site.connected, as shown. above. I also think it's nice to have Validator::Site.connected in each test case, to more clearly show what's going on in the test.
> 
> OK. Should be add a separate pull request for the new version of request_status_and_confirm() before finishing this PR?

Yes, let me add a PR for that first.

emiltin:
done

PR #235, opened 2023-04-14
Add test case for M0023 timeout of dynamic bands

otterdahl:
The M0023 command was added in SXL 1.1. This PR adds a test case for this command

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -32,7 +32,7 @@ GEM
     nio4r (2.5.9)
     rake (13.0.6)
     regexp_parser (2.7.0)
-    rsmp (0.17.2)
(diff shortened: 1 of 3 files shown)


otterdahl:
The Gem TLC doesn't seem to support M0023 yet

emiltin:
> The Gem TLC doesn't seem to support M0023 yet

right

emiltin:
I added support for M0023 in the gem tlc. It's simplistic, and is not yet actually used in the dynamic bands. But it makes the test pass.

I added a check for setting the timeout to 10 (in addtional to the existing timeout of 0)

emiltin:
rebased on master

otterdahl:
Looks good

PR #236, opened 2023-04-14
activate inputs for raising alarms, and retire scripts

emiltin:
Removes all support for external scripts as a way of activating alarms. Insteed we now rely on acticating inputs. 
Relevant tests have been refactored and now use the helper with_alarm_activated().

### DIFF
--- a/.github/workflows/gem_supervisor.yml
+++ b/.github/workflows/gem_supervisor.yml
@@ -30,7 +30,7 @@ jobs:
 
     - name: Run tests
       shell: bash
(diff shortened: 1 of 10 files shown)


emiltin:
should fix #22.
@otterdahl do you agree on retiring scripts, and relying only on input activation?

otterdahl:
Yes, it's OK. Scripts was a temporary workaround and was prone to failure. The only downside to using inputs is that is requires a special configuration of the TLC, but I think we can live with that.

PR #237, opened 2023-04-14
Add test case for S0032 coordinated control

otterdahl:
S0032 coordinated control was added in SXL 1.1. This PR adds a test case for it


otterdahl:
The tests seem to pass. Merge?

PR #239, opened 2023-04-28
fix workflow test selection

emiltin:


emiltin:
for some reason, workflow files for physical tlc's was running only an alarm test. this pr fixes it.

it also removes the --tag "~script" from the gem tests. we now rely only on activation inputs for triggering alarms, which means all tests can be run on the gem TLC emulator. There's a test that take 1min, because it tests a timeout. I've choosen to run this as well, so we run every tests.

PR #243, opened 2023-05-04
Docker doc

emiltin:


emiltin:
fixed, but GH pages does not update the site before the gh-pages branch is updated. maybe look into the new option to build gh pages from an action.

emiltin:
The best would be to have a separate gh-pages site to push when a PR is opened. It would act as a staging server, so we can view how the updates docs would look when published with gh-pages. 

A git repo can only have one gh-pages site, but it possible to keep a separate github repo that acts as a staging server, and use workflow dispatch to update the staging server gh-pages site whenever the PR is updated.

What we've lived with so far is simply that our site is only updated when we psuh to master. So we have no way of checking the result beforehand.

emiltin:
for now i'll simply update docs. we can improve work on a staging server for our doc site later if neeeded

emiltin:
closing this PR, will reopen one form the docs branch, so that docs are rebuild

PR #246, opened 2023-05-09
how to run with docker

emiltin:
Adds documenation about how to run with Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM ruby:latest
-COPY . .
+COPY .ruby-version .rspec Gemfile Gemfile.lock LICENSE ./
+COPY spec spec
 RUN bundle install
-EXPOSE 12111
-ENTRYPOINT [ "bundle", "exec", "rspec" ]
-CMD [ "spec/site/core", "spec/site/tlc" ]
+EXPOSE 13111
+ENTRYPOINT [ "bundle", "exec", "rspec" ]		# default executable
+CMD [ "spec/site/core", "spec/site/tlc" ]		# default arguments
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,4 +3,4 @@ COPY . .
 RUN bundle install
 EXPOSE 12111
 ENTRYPOINT [ "bundle", "exec", "rspec" ]
-CMD [ "spec/site" ]
+CMD [ "spec/site/core", "spec/site/tlc" ]
--- a/docs/pages/configuring.md
+++ b/docs/pages/configuring.md
@@ -3,7 +3,7 @@ layout: page
 title: Configuration
 permalink: /config/
 parent: Getting Started
-nav_order: 2
+nav_order: 3
 ---
 
 # Configuration
--- a/docs/pages/docker.md
+++ b/docs/pages/docker.md
@@ -1,0 +1,45 @@
+---
+layout: page
+title: Docker
+permalink: /docker/
+parent: Getting Started
+nav_order: 1
+---
+
+# Running with Docker
+This section explains how to run the RSMP Validator using Docker. You can also [install locally]({{ site.baseurl}}{% link pages/installing.md %}).
+
+## Setup
+1. Download and install [Docker](https://www.docker.com).
+
+2. Download the Docker image with `docker pull ghcr.io/rsmp-nordic/rsmp_validator:docker`. You can find the [available docker images on github](https://github.com/rsmp-nordic/rsmp_validator/pkgs/container/rsmp_validator).
+
+
+2. Create a local folder for storing validator configurations. It will serve the same purpose as the [config folder](https://rsmp-nordic.org/rsmp_validator/config/) when you install manually.
+
+3. Create a configuration file for the equipment, simulator or supervisor that you want to test. Check out an [example for TLC emulator](https://github.com/rsmp-nordic/rsmp_validator/blob/master/config/gem_tlc.yaml) build into the the rsmp Ruby gem, or ream more about [site test configurations](https://rsmp-nordic.org/rsmp_validator/config/#options-for-site-testing) and [supervisor test configurations](https://rsmp-nordic.org/rsmp_validator/config/#options-for-supervisor-testing).
+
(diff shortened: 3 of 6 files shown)


emiltin:
@otterdahl that do you think? https://rsmp-nordic.org/rsmp_validator/docker/

otterdahl:
It looks good!

I just noticed small typos:
* `docker pull ghcr.io/rsmp-nordic/rsmp_validator:docker` should be `docker pull ghcr.io/rsmp-nordic/rsmp_validator:master`.

(Where `master` refers to the master branch. If we choose to tag and make a release, this label can be used to refer to that tag)

emiltin:
> * `docker pull ghcr.io/rsmp-nordic/rsmp_validator:docker` should be `docker pull ghcr.io/rsmp-nordic/rsmp_validator:master`.

fixed

emiltin:
I think we should add this:

```
The default that the validator will run is determined by the CMD directive in the Dockerfile. We can update this if needed, but it is easy to override this as described in the doc. (ENTRYPOINT is always run)

ENTRYPOINT [ "bundle", "exec", "rspec" ]
CMD [ "spec/site/core spec/site/tlc" ]
```

And use `spec/site/core spec/site/tlc`to be specific.

BTW, since there's a Dockerfile

otterdahl:
I think it needs to be written like this:
`CMD [ "spec/site/core", "spec/site/tlc" ]`

emiltin:
Running on the terminal fails with:

`terminal
code/rsmp_validator (docs) % docker container run -it -v $PWD/config/private -p 12111:12111 rsmp_validator                     
Unable to find image 'rsmp_validator:latest' locally
docker: Error response from daemon: pull access denied for rsmp_validator, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
`

I think the issue is that docker uses `latest`as the default tag name, while the image uses `master`. I really think we should use `latest` as the tag when publishing the (latest) master branch.

otterdahl:
I think 'master' as a tag is what the official github action for docker uses as default.

emiltin:
From the docs at https://github.com/docker/metadata-action#latest-tag, it seems a `latest` tag is automatically generated, but only for [type=ref,event=tag](https://github.com/rsmp-nordic/rsmp_validator/pull/246#typeref), which is when you push a git tag. We're not (yet) using git tags in the validator, so it's never generated.

I'll try their suggestion to conditionally tag with latest for the default branch.

emiltin:
I've looked at bit at this today. I still have several issue when trying to run it via Docker.

I suggest updating the Dockerfile:

```
FROM ruby:latest
COPY .ruby-version .rspec Gemfile Gemfile.lock LICENSE ./
COPY spec spec
RUN bundle install
EXPOSE 13111
ENTRYPOINT [ "bundle", "exec", "rspec" ]		# default executable
CMD [ "spec/site/core", "spec/site/tlc" ]		# default arguments
```

1. Only necessary files are copied
2. EXPOSE is changed from 12111 to 13111, which is what the validator actually listens on. Note that EXPOSE is really just documentation, as you still need to expose ports. when you run the container  with the `-p 13111:13111`. This is why it worked even with 12111.

I can build the image locally (from the local git repo folder) with:

```shell
% docker build . --tag rsmp_validator
```

And then run with default args:

```shell
docker run --rm --name rsmp_validator -it -v $PWD/config:/config -p 13111:13111 rsmp_validator
```

Here I'm mapping the local folder ./config to ./config in the container.  When you do this in the checkout repo folder, it means that the standard valdiator.yaml is read, which works for testing the gem TLC.

Importantly, port 13111 must be mapped, not 12111.

Using `--rm` means the container will be remove after running, which is good in case you want to run it again. Otherwise you can an error next time you run the same command, because a container isalready exists with the name rsmp_validator.

Or you can run with custom args:

```shell
docker run --rm --name rsmp_validator -it -v $PWD/config:/config -p 13111:13111 rsmp_validator spec/site/tlc --format Validator::Brief --tag "~@slow"
```


You can also persist additinal log output to a file in a separate mounted log folder:
```shell
docker run --rm --name rsmp_validator -it -v $PWD/config:/config -v $PWD/log:/log -p 13111:13111 rsmp_validator spec/site/core --format Validator::Brief --format Validator::Details --out log/validation.log
```

After publishing the image with the `latest` tag also, you can run with just:
```shell
docker run --rm --name rsmp_validator -it -v $PWD/config:/config -p 13111:13111 ghcr.io/rsmp-nordic/rsmp_validator
```

There's no need to pull the image first, as run will do it if neeed.

I'll update the docs.

otterdahl:
> Importantly, port 13111 must be mapped, not 12111.

Yeah, It depends on the validator.yaml (or what config file it points to). I've only used 12111 in my configuration. Feel free to change so it stays consistent and matches the documentation.

> Seems like we somehow need to add `linux/arm64/v8`as a platform when building the image in the github action?

There is some documentation at https://docs.docker.com/build/ci/github-actions/multi-platform/ but I can't find any information about linux/arm64/v8.

> Using `--rm` means the container will be removed after running, which is good in case you want to run it again. Otherwise you get an error next time you run the same command, because a container already exists with the same name.

I don't think `--rm` is really required. I can run the container again and again without removing the container. (using `docker start` instead of `docker run`). Keeping the container between runs means that any files modified in the container stays intact - which may or may not be what you want. However, we kind of assume that anyone using docker have some basic skills how to run docker, so I don't think we really need to include this.

emiltin:
> I don't think --rm is really required. I can run the container again and again without removing the container. (using docker start instead of docker run). Keeping the container between runs means that any files modified in the container stays intact - which may or may not be what you want. However, we kind of assume that anyone using docker have some basic skills how to run docker, so I don't think we really need to include this in the documentation.

ok i removed it from the examples.

emiltin:
fixed the warning on Mac with a multi-arch docker image build. i think this is ready for merging?
updated docs can be viewed already at https://rsmp-nordic.org/rsmp_validator/docker/. because we don't yet have staging server, the docs are updated as soon as you push to the docs branch, which this PR uses

otterdahl:
Looks good

PR #253, opened 2023-05-15
Fix timeout for test cases for invalid status

otterdahl:
The test cases for invalid status seems to use the timeout for command response. Most likely, it should use the timeout for status response instead.


otterdahl:
Added e902a9ea20adcf63eee0d97249b4440bc9374c4f too by mistake

emiltin:
Merging.
PS: You can use `rebase -i master` to remove/squash/rename commits.

PR #256, opened 2023-05-17
Alarm component, fixes #247

### LINKED ISSUE #247: 'A0301 can be acknowledged when input is activated' uses main_component ID, but should use detector logic component ID
According to the SXL specification (1.0.14 through 1.1) alarm A0301 belongs to the "Detector Logic" object type, but the test is using the main_component ID when sending an alarm acknowledge.

emiltin:
Uses component in alarm tests.

Note that the config for when you run a gem TLC site changed slightly, so that the component can be specified: 
https://github.com/rsmp-nordic/rsmp/pull/48

The test config format also changed slight, so that you can specify the componet:

```yaml
alarm_activation:
  A0301: 
    input: 152
    component: KK+AG9998=001DL001
```

This means that exisist configs must be updated.


emiltin:
One of the tests uses A0302, I changed this to A301 - is this OK? It did seem like there was any real reason for using A302, so for the time being I think it's easier to use the same alarm for all tests, so that the configs are simpler.

otterdahl:
> One of the tests uses A0302, I changed this to A301 - is this OK? It did seem like there was any real reason for using A302, so for the time being I think it's easier to use the same alarm for all tests, so that the configs are simpler.

The reason we used A0302 (and not A0301) is because A0302 is a logic error (and not hardware). This means it is  easier to trigger it though software.

As the description for A0302 states, this alarm triggers if (a detector is continuously on or off during an extended time. If this condition is met, this is triggers an error. Setting Input and using a special configuration in the TLC, we can change this time and trigger the alarm after a very short delay.

We haven't figured out a way to trigger hardware alarms - unless the TLC has some special ability to simulate those.

emiltin:
So should we use A0302 for all the tests?

otterdahl:
> So should we use A0302 for all the tests?

Yeah, we might as well do that.

emiltin:
Ok. The test 'Clock is used for alarm timestamp' also uses A0301. Guess we change that to A0302 as well?

otterdahl:
> Ok. The test 'Clock is used for alarm timestamp' also uses A0301. Guess we change that to A0302 as well?

Yes I think so. It makes it more consistent

PR #272, opened 2023-05-25
merge #267 (fix itc2 workflow) to master

emiltin:
#267 fixed the itc-2 workflow. a PR against the equipment triggers the EC-2, ITC-2 and ITC-3 tests, but when merged, the changes of course go to the equipment branch, so they are not yet in the master branch. the daily test runs are done from master, so we need to merge equipment to master, which this pr does.

### DIFF
--- a/.github/workflows/swarco_itc2.yml
+++ b/.github/workflows/swarco_itc2.yml
@@ -18,12 +18,10 @@ jobs:
 
       - name: Install wine32
         run: |
+          sudo rm -f /etc/apt/sources.list.d/microsoft-prod.list
           sudo dpkg --add-architecture i386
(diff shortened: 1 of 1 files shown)


emiltin:
It's a bit inconvenient to have to first do a PR against equipment, then merge to master.

As an alternative, do could perhaps trigger the equipment tests when when a specific label is set. then we could do a PR against master (or any other branch if needed), and just add a `equipment` label when we want equipment tests to run?

otterdahl:
Yeah. Using labels seems like a good idea. In that case, we don't need `equipment` as a branch, only as a `label` on the PR. Let's try that

PR #312, opened 2023-05-30
trigger workflow by label

emiltin:


emiltin:
Seems to work. Equipment tests start when you add the 'equipment' label.

emiltin:
So the workflow now is that you can create a PR against master (or any other branch). If you want equipment tests (ec2, itc3, itc3) to run, just add the 'equipment' label.

PR #326, opened 2023-06-22
update rsmp, to use async 2

emiltin:


emiltin:
hi @otterdahl this PR pulls in the latest rsmp gem, which upgrades the async dependciy to version 2. this changes means we no longer support older Ruby version 3.0 o 3.1, only the newer 3.2 series. just a heads up in case it needs upgrading the our test hub PCs.

Issue #354, opened 2023-11-30
modes_spec: "yellow flash affects all signal groups" has a hardcoded timeout of 10 seconds


sveitech:
```
 # Verify that we can yellow flash causes all groups to go to state 'c'
    #
    # 1. Given the site is connected
    # 2. Send the control command to switch to Yellow flash
    # 3. Wait for all groups to go to group 'c'
    # 4. Send command to switch to normal control
    # 5. Wait for all groups to switch do something else that 'c'
    specify 'yellow flash affects all signal groups', sxl: '>=1.0.7' do |example|
      Validator::Site.connected do |task,supervisor,site|
        prepare task, site
        timeout =  10 

        switch_yellow_flash
        wait_for_groups 'c', timeout: timeout      # c mean s yellow flash

        switch_normal_control
        wait_for_groups '[^c]', timeout: timeout   # not c, ie. not yellow flash
      end
    end
```

This timeout of 10 seconds should be configurable in the config file. The timeout is also hardcoded in the "wait_for_groups" function, even though it is passed as a parameter.

```
def wait_for_groups state, timeout:
    timeout = 10
    regex = /^#{state}+$/
    wait_for_status(@task,
      "Wait for all groups to go to yellow flash",
      [{'sCI'=>'S0001','n'=>'signalgroupstatus','s'=>regex}],
      update_rate: 0,
      timeout: timeout
    )
  end
```

emiltin:
Fixed with #359

PR #359, opened 2023-12-13
yellow flash timeout config

emiltin:


emiltin:
Fixes #354. 
A timeout of 60 is used in the test configs for EC-2, ITC-2 and ITC-3. This can be adjusted if needed.

PR #364, opened 2024-01-12
Validator.get_config helper to fetch config

emiltin:
Fetches a config (optionally using a default value). Raise an error if not found. 
This PR updates all tests to use this helper.

Using this helper will stop a test with a clear error message if a config is missing. Otherwise a test might run but fail in weird ways.


emiltin:
Note that if a test setup was missing some configuration, and nil was somehow working as a default, running the test will not abort with a notice that the config is missing. In that case, just add the missing config to your test config.

PR #379, opened 2024-04-03
update actions/upload-artifact and actions/checkout to @v4

emiltin:


emiltin:
required since github is transitioning Node.js from 16 to 20.

PR #380, opened 2024-05-16
Update the workflow for the new ITC-3 action runner

otterdahl:
The workflow needs to updated in order to support the new ITC-3 action runner


otterdahl:
The test fails with the Ruby installation using by the runner. In case ruby is already installed on the runner this step is not necessary. I'll try to remove this step and see if it works.

otterdahl:
Ruby is not installed by default, so I'll add the ruby setup step again

PR #395, opened 2024-06-19
Try the testhub label

emiltin:


emiltin:
@otterdahl see https://github.com/rsmp-nordic/rsmp_validator/discussions/396

PR #420, opened 2024-09-02
Create tecsen_tmacs_supervisor.yaml

mariusrazvanplugaru:


emiltin:
Thank you for this config file. Looks good, should be useable for testing your TMAC supervisor.
Just keep in mind that the test set for supervisors is very limited at the moment.
I'll add github workflow for TMAC.

PR #422, opened 2024-09-03
run gem tests on matrix of core versions

emiltin:


emiltin:
Currently the --core option is used when starting the Gem TLC, but this should be changed so that we instead tells the supervisor (validator to only accept a particular core version.

emiltin:
Them gem tlc and gem supervisor tests now use a matrix of core version, and also still a matrix of OS'es.
To avoid a n*m explosion of jobs, each has been split into two workflows:
- one that runs on a matrix of OS'es with the latest core version
- one that runs on a matrix of core versions, on ubuntu-latests

I think this is sufficient.

Uploading of log files has also been modernized and uses artifact names line `validator-gem-tlc-ubuntu-latest-core-3.1.2-run-10718203050-64-1.log`. Instead of the date, the github workflow id, run and attempt ids are used.

emiltin:
Running all the tests on different core version surfaced a handful of failing tests cases. But it was all a matter of adding rspec tags to restrict the core versions that the tests should be run on. For example, S0035 uses the new array data type which was introduced in core 3.2. So tests involving S0035 can only be expected to succeed when core >= 3.2, and have therefore been tagged accordigly.

emiltin:
The `restrict_testing` config has been changed bit. The core version is not included under this key anymore, instead it's simply an option at the root level called 'core_version'. If you set this, then the validator will force the core version during the version handshake.

You can also force the core version by setting the env variable CORE_VERSION. When running on a matrix of core version, the workflow uses this mechanism to force the core version for each job.

PR #423, opened 2024-09-04
Update semaforica_cartesio.yaml

mariusrazvanplugaru:
Change address port


emiltin:
I would prefer to change the port of TMacs to keep all TCLS using the same port, just for consistency.

mariusrazvanplugaru:
ok how can i close this? for open a new one for TMacs

PR #431, opened 2024-09-09
run itc3 and cartesio on a matrix of core verison, but just with one …

emiltin:
…version

### DIFF
--- a/.github/workflows/gem_supervisor.yaml
+++ b/.github/workflows/gem_supervisor.yaml
@@ -9,11 +9,11 @@ on: push
 jobs:
   validator:
     timeout-minutes: 10
+    runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
         core: ['3.1.2', '3.1.3', '3.1.4', '3.1.5', '3.2', '3.2.1', '3.2.2']
-    runs-on: ubuntu-latest
+       fail-fast: false
 
     steps:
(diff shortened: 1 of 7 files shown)


emiltin:
@mariusrazvanplugaru woud you like to test your cartesio with different core version? curently it reports that it suports only the latest 3.2.2.

mariusrazvanplugaru:
Sorry but this will be visible in Compliance page? Actually we can choose manually with one configuration the version of Core, but is possible to change that to be made automatically.

emiltin:
Not yet, but  we plan to expand how results are shown. At the moment, the flag is green if all tests are green, which would mean if tests pass on all core versions.

emiltin:
Do you mean the TLC always uses 3.2.2, regardless of the Version messages exchanged?

mariusrazvanplugaru:
> Not yet, but we plan to expand how results are shown. At the moment, the flag is green if all tests are green, which would mean if tests pass on all core versions.

Ok, it will be great. When you have that we could configure on controller to accept a matrix of versions and we will start to test a set of core versions.

emiltin:
Alright. The workflow already uses a matrix, although only 3.2.2 is currently in the list of core versions to test. So when you like you can do PR with changes at https://github.com/rsmp-nordic/rsmp_validator/blob/main/.github/workflows/semaforica_cartesio.yaml#L22

PR #460, opened 2024-11-19
add workflow for cross-rs4s

emiltin:


emiltin:
Hi @pksoft72, the runner cannot install Ruby automatically because the setup-ruby GitHub Action onyl supports Ubuntu, and you're running on Debian:
```
Using 3.3 as input from file .tool-versions
Error: The current runner (debian-12-x64) was detected as self-hosted because the platform does not match a GitHub-hosted runner image (or that image is deprecated and no longer supported).
In such a case, you should install Ruby in the $RUNNER_TOOL_CACHE yourself, for example using https://github.com/rbenv/ruby-build
You can take inspiration from this workflow for more details: https://github.com/ruby/ruby-builder/blob/master/.github/workflows/build.yml
$ ruby-build 3.3 /home/rsmp-testing/actions-runner/_work/_tool/Ruby/3.3/x64
Once that completes successfully, mark it as complete with:
$ touch /home/rsmp-testing/actions-runner/_work/_tool/Ruby/3.3/x64.complete
It is your responsibility to ensure installing Ruby like that is not done in parallel.
```

setup-ruby documentation:
https://github.com/ruby/setup-ruby

You can either change to a Ubuntu system, Ruby is then automatically installed, and later updated when needed. This is probably what I would recommend.

Or you can manually install Ruby 3.3.6, but might have to repeat whenever Ruby updates.

pksoft72:
problems with ruby should be solved

pksoft72:
Run again, please. I had to restart actions runner to some environment changes take effect.
It should give some result - at least like my test on cloned repository:
https://github.com/pksoft72/rsmp_validator/actions/runs/11914592499/job/33202888028

emiltin:
So

> Run again, please. I had to restart actions runner to some environment changes take effect. It should give some result - at least like my test on cloned repository: https://github.com/pksoft72/rsmp_validator/actions/runs/11914592499/job/33202888028

Sorry,  still the same error:
Error: The current runner (debian-12-x64) was detected as self-hosted because the platform does not match a GitHub-hosted runner image (or that image is deprecated and no longer supported).

pksoft72:
#461 needs testhub label and then it should look like: 
https://github.com/pksoft72/rsmp_validator/actions/runs/11939581085/job/33280406728

pksoft72:
I will work on fixes.

for example:
cycle time is set with M0018 - Failed: command response 732209ed was rejected with 'Plan length can be only extended but cannot be shrinked (must be >= 91) !'

Cycle time is set to 5 seconds. 
Our controller can only extend plan and not shrink.
I can prepare special unrealistic very short plan for testing. But I prefer testing in "near real conditions".
Or I can prepare PR for extending configuration of this fixed value, so Cross controller will have some big value here.

What would you prefer?

emiltin:
Would you mind opening an issue for each failing tests that you would like to discuss? That will make it easiser for us navigate, I think.

PR #461, opened 2024-11-20
add workflow for cross_rs4.yaml (2)

pksoft72:
based on  "add workflow for cross-rs4s #460" by Emil Tin 
fixed to ruby version 3.3.6 - as easiest way to make it run on Debian 12


pksoft72:
need "testhub" label

emiltin:
Interesting, do you know you it does not use to .tool-versions repo file on Debian?

pksoft72:
It is probably using .tool-versions, but I had problems installing 3.3, as ruby-build --list offered only 3.3.6 and action runner insisted on 3.3. 
I am beginner in ruby so I am happy I can understand code in tests.

Issue #462, opened 2024-11-21
M0018 set cycle time test - configurable cycle length


pksoft72:
In this test

```ruby
    # 1. Verify connection
    # 2. Send control command to set cycle time
    # 3. Wait for status = true
    it 'M0018 set cycle time', sxl: '>=1.0.13' do |example|
      Validator::Site.connected do |task,supervisor,site|
        status = 5     ## <<<<<<<<< here
        plan = 0
        prepare task, site
        set_cycle_time status, plan
      end
    end

```

is hardcoded cycle length value to 5 seconds.

I prefer testing on near real conditions and our Cross RS4S controller supports only extending original plan and not shrinking.

- I can make some special very short plan to pass this test or 
- I can modify code to use configuration here with default value 5.

emiltin:
I don't think we have any requirement that the  cycle time can be reduced. Perhaps we should change the test to first read the cycle time, then extend it, e.g. with 5s.

pksoft72:
and then return back, otherwise it will grow with each test run

emiltin:
You right, we should take care to reset it. Another option would be to keep the normal and extended cycle time as a test config, but I think it's easier if we can avoid adding test configs.

emiltin:
The M0018 tests now extends the cycle time by 5s and reverts afterwards.

PR #467, opened 2024-12-11
M0018 cycle time

emiltin:
Improves the M0018 set cycle time test. It now reads the current time, extend with 5 seconds, then resets back to the previous cycle time.

The rsmp gem TCL emulator was updated to implement M0018 to pass the test.

To support this, a few other things were changed:

- gems updated
- 'restrict_testing' replaced by 'sxl_version' in configs
- 'rsmp_versions' replaced by 'core_version' in configs, the value 'all' is not allowed anymore
- signal plans are now defined in the config as:
 ```yaml 
  signal_plans:
    1:
      cycle_time: 6
      states:
        A1: '111NBB'
        A2: '11NBBB'
        B1: 'BBB11N'
        B2: 'BBB1NB'
      dynamic_bands:
        1: 0
        2: 5
  ```
- the the cycle time is now part of signal plans, rather than a global settings in the main component.
- how to raise alarms is now defined in the config under inputs, rather than under alarms, e.g.:
  ```yaml
  inputs:
    total: 8
    programming:
      7:
        raise_alarm: A0302
        component: DL1
  ```
- new Validator::ProgrammingHelpers to read alarm activation settings from the config
- updated with_alarm_activated helper
- new read_cycle_times and read_current_plan helper methods

### DIFF
--- a/config/cross_rs4s.yaml
+++ b/config/cross_rs4s.yaml
@@ -82,10 +82,6 @@ items:
   force_input: 5
   outputs: [1,3,8]
 startup_sequence: 'efg'
-#alarm_activation:       # how to trigger alarms by forcing inputs
-#  A0303:                # alarm A0302
-#    input: 7            # can be triggered by forcing input 8
-#    component: DL1      # and will active on component DL1
 alarms:
   A0302: 
     activation_input: 1


emiltin:
Should fix #462

PR #472, opened 2024-12-17
Update kapsch_etx.yaml

emiltin:


anezsergi:
@emiltin Ok, it seems that the OS version is too old. We'll have to wait until tomorrow and I'll try another computer. I don't have access right now. 

Will I need a new token or is it possible to copy files to a new computer?

emiltin:
> @emiltin Ok, it seems that the OS version is too old. We'll have to wait until tomorrow and I'll try another computer. I don't have access right now.
> 
> Will I need a new token or is it possible to copy files to a new computer?

I don't know if it's possible to copy files. I would reinstall with a new token.

anezsergi:
> > @emiltin Ok, it seems that the OS version is too old. We'll have to wait until tomorrow and I'll try another computer. I don't have access right now.
> > Will I need a new token or is it possible to copy files to a new computer?
> 
> I don't know if it's possible to copy files. I would reinstall with a new token.

Hi @emiltin , I need a token.  when you can send me one, please. the last one I have had some problems and the token has expired I think

emiltin:
Hi @anezsergi did you manage to install the runner on a newer machine/server/container?

anezsergi:
@emiltin 
Yes, I reinstalled runner in new server and it seems run correctly, can you confirm it?

emiltin:
You runner don't have permissions to create  /opt/hostedtoolcache
https://github.com/rsmp-nordic/rsmp_validator/actions/runs/12396523463/job/34648779378?pr=468


See https://github.com/actions/runner/issues/2892
You might just need to create the folder /opt/hostedtoolcache and have it chown'ed by the user that does the running.

anezsergi:
Thanks  @emiltin 
I just did it.

emiltin:
It's now running and communicating with your controller.
The controller is reporting that it supports core versions  [3.1.5, 3.2, 3.2.1], is this correct?

anezsergi:
No, I want to change this config but I see it was working and i will change it when finish test.

emiltin:
> No, I want to change this config but I see it was working and i will change it when finish test.

Ok, I can see it's busy with the 3.2.1 test.

anezsergi:
> > No, I want to change this config but I see it was working and i will change it when finish test.
> 
> Ok, I can see it's busy with the 3.2.1 test.

Is this normal? It's been going on for 24 minutes and it hasn't finished yet

emiltin:
> Is this normal? It's been going on for 24 minutes and it hasn't finished yet

It depends a lot of the test config and the controller. You have set some quite long timeouts in your test config of several minutes, e.g: 
https://github.com/rsmp-nordic/rsmp_validator/blob/main/config/kapsch_etx.yaml#L22

This mean tests might wait for 180 or 240 seconds before failing. Since there are many tests, it could easily sum up. So generally, try to reduce the timeouts in the config to as low as possible.

anezsergi:
Thanks @emiltin, I'll review all these timeouts this afternoon.

emiltin:
Unfortunately, you can't really see the logs on github actions as they are being generated, so until the test completes I don't have any knowledge about why it's taking so long. 

Another thing that can make testing take longer is if you specify a lot of objects to test, under the 'items' key in the config.
For example, you have 10 inputs at https://github.com/rsmp-nordic/rsmp_validator/blob/main/config/kapsch_etx.yaml#L85. This is good, as your a thorough, but it's ok to just list two. Note that you might list two, even though you have more inputs configured in the device.

anezsergi:
Thanks @emiltin, I'll review all these waiting times this afternoon and items config.

the mechanism to do this modified the file and created a PR?

emiltin:
> Thanks @emiltin, I'll review all these waiting times this afternoon and items config.
> 
> the mechanism to do this modified the file and created a PR?

Yes, but an admin (e.g. me) must approve running tests on PRs from other repos. I'm leaving today to holidays. If you create one now (even with just placeholder changes) I can approve it.

emiltin:
Of course you should also check how your controllers handles the different commands, a timeout only occurs when the controller does not reply in time.

The tests will run nightly, so you can always work on your implementation and check the results of these nightly runs without my involvement at https://github.com/rsmp-nordic/rsmp_validator/actions/workflows/kapsch_etx.yaml

anezsergi:
Ok @emiltin I just changed some parameters in the yaml file (create a PR) and I think this way I will be able to work better before holidays.

emiltin:
This run completed in about 2 hours, with some errors:
https://github.com/rsmp-nordic/rsmp_validator/actions/runs/12377309481/job/34654744271?pr=472

I cancelled the other runs in this PR.

PR #473, opened 2024-12-18
Time plan nrs

emiltin:


emiltin:
Some tests are failing on Windows.

Basically, the async gem, and the Ruby Fiber Scheduler more generally, which we rely heavily on for concurrency, is not fully supported yet on Windows, see e.g. https://github.com/socketry/async-http/issues/163. For whatever reason it seemed to work before the recent and ruby updates.

I believe the best thing to do is to remove Windows from our OS matrix, since it's not expected to work. I think almost nobody is using the Validator on Windows anyways.

Tests are also very slow on Windows, because the ruby/setup-ruby GitHub Actions installs Ruby in a few seconds on Mac and Linux, but it takes minutes on Windows, see https://github.com/ruby/setup-ruby/discussions/635#discussioncomment-11333609.

emiltin:
I tried to run `pcregrep --color='auto' -n "[\x80-\xFF]" **/**" to check for non-ascii characters. I did not find any in the spec/ of config/ folders. But there's a number in the YAMl files in the schema repo, typically _curly quotes_.

PR #475, opened 2024-12-19
Update kapsch_etx.yaml file

anezsergi:


emiltin:
The jobs in this PR is now running, with your changed to the config file.

emiltin:
This error "Address already in use - bind(2) for 0.0.0.0:12111" indicates that a "rsmp supervisor" process is still running on the runner machine.

emiltin:
I'm going to merge this, so the updated config will be used for the nightly builds.

emiltin:
Test runs now complete in around 25 min. Merging.

emiltin:
The problem still seems to be a left-over process that is using port 12111.
```
Address already in use - bind(2) for 0.0.0.0:12111
```

Issue #487, opened 2025-01-07
Kapsch ETX acknowledgement timeouts


emiltin:
The test config uses a short acknowledgement timeout of just 1s. 

But when testing, there are many case where an acknowledgement is not received within 1s:

```
    2025-01-07T12:40:41.751Z  51584  KTT+TC=101TC001      KTT+TC=101TC001      dd52  Sent CommandRequest  {"mType":"rSMsg","type":"CommandRequest","cId":"KTT+TC=101TC001","arg":[{"cCI":"M0104","cO":"setDate","n":"securityCode","v":"1111"},{"cCI":"M0104","cO":"setDate","n":"year","v":"2020"},{"cCI":"M0104","cO":"setDate","n":"month","v":"9"},{"cCI":"M0104","cO":"setDate","n":"day","v":"29"},{"cCI":"M0104","cO":"setDate","n":"hour","v":"17"},{"cCI":"M0104","cO":"setDate","n":"minute","v":"29"},{"cCI":"M0104","cO":"setDate","n":"second","v":"51"}],"mId":"dd52b832-148f-4ac1-be8d-d4175f6752ee","ntsOId":"","xNId":""}
    2025-01-07T12:40:43.440Z  51584  KTT+TC=101TC001      -                    -     No acknowledgements for CommandRequest dd52 within 1 seconds
```

Either the TLC must send all acks within 1s, or the timeout config must be increased.

emiltin:
This is causing a large number of tests to fail.

anezsergi:
I have just made a pull request changing timeouts.

emiltin:
Thanks, I merged it.

emiltin:
Tests are now running. This should also show the result of your other fixes.

PR #491, opened 2025-01-08
increase etx ack timeout from 1 to 10s

emiltin:


emiltin:
Let's see if this fixes #487.

emiltin:
Outdated.

PR #495, opened 2025-01-09
Store tests results as json artefact (for the rsmp gem workflow)

emiltin:
Adds a json output when running the rspec command. The resulting file is then uploaded as an artefact.


emiltin:
Next step is to associate the output artefacts with the core and sxl version used when testing, so the doc generation can show which versions were tested.

Currently the results is saved to a file with names like:
```
results-gem-tlc-ubuntu-latest-core-3.1.3-run-12689397028-177-1
results-gem-tlc-ubuntu-latest-core-3.2.2-run-12689393344-132-1
```

This includes the core version, but not the sxl. The reason is that the core version is set by the github action matrix, but the sxl is set in the test config. We could perhaps read the sxl from the config and incude it in the artefact name.

But it might not be convenient or even possible to list available artefact, so we might need another way to directly list the core and sxl verisons tested. Maybe a separate action step can handle this and create a artefact that lists the core/sxl versions tested.

emiltin:
It's possible to output annotations. See e.g. https://github.com/Drieam/rspec-github.
Should check if that's a better way to output results.

emiltin:
done, we're now have a flow where tests results are stored as artifacts, and a separate workflow ingest these, computes stats and generate static pages at https://rsmp-nordic.org/compliance/

PR #496, opened 2025-01-15
cross_rs4s configuration changes

pksoft72:


emiltin:
What's the reason for changing the name of the runner in the workflow file? Currently the workflow will not since the name does not match the tag set on the runner here on github.

pksoft72:
It should have been another commit :-( 
[b495600](https://github.com/rsmp-nordic/rsmp_validator/pull/496/commits/b495600c693d3ce0a52c41d5562698457cf20818)

Issue #465, opened 2024-11-22
Test of flashing yellow should accept dark on some signal groups


pksoft72:
This test switches to flashing yellow and expect all groups to have state 'c'
```ruby
    # Verify that we can yellow flash causes all groups to go to state 'c'
    #
    # 1. Given the site is connected
    # 2. Send the control command to switch to Yellow flash
    # 3. Wait for all groups to go to group 'c'
    # 4. Send command to switch to normal control
    # 5. Wait for all groups to switch do something else that 'c'
    specify 'yellow flash affects all signal groups', sxl: '>=1.0.7' do |example|
      Validator::Site.connected do |task,supervisor,site|
        prepare task, site
        timeout =  Validator.get_config('timeouts','yellow_flash')

        switch_yellow_flash
        wait_for_groups 'c', timeout: timeout      # c mean s yellow flash

        switch_normal_control
        wait_for_groups '[^c]', timeout: timeout   # not c, ie. not yellow flash
      end
    end

```
According to https://rsmp-nordic.github.io/rsmp_sxl_traffic_lights/1.2.1/signal_group_status.html, 'c' means "Manual control to flashing yellow" and output is "Yellow flash".

But there are signal groups - typically pedestrian or any single color groups - which are dark in yellow flash mode so I suggest to report 'a' = Disabled/dark for these groups and accept such groups in this test.

So I suggest change 1st wait_for_groups to

```ruby
wait_for_groups '([a,c]*c[a,c]*)', timeout: timeout      # c mean s yellow flash, a means dark
```

I could pass this test by removing all non R-Y-G signal groups, but I prefer real conditions.

SwarcoPalm:
These are logical states, startup e,f,g does not tell you about the physical output, startup sequence is country dependent. This is how it is supposed to be

otterdahl:
@pksoft72 Thank you for the feedback. I think you make a good point. However, as @SwarcoPalm points out, these are internal states and doesn't necessarily reflect the output. Also, we rely on the internal states in order to tell whether the command has been successful or not. If we were to accept that some signal groups goes to dark mode instead we also somehow need to specify the type of signal group (pedestrian, bicycle, vehicle) which would the configuration more complex and error prone. I'm not convinced that it is worth the trouble.

pksoft72:
Ok, I have probably wrongly interpreted column Output in table
[...]

PR #497, opened 2025-01-16
cross_rs4s configuration changes

pksoft72:


.github/workflows/cross_rs4s.yaml
```diff
@@ -6,20 +6,25 @@
...
+  workflow_dispatch:  # should enable manual start of test
   pull_request:
     types: [ opened, reopened, synchronize, labeled ]
   schedule:
     # schedule runs only on the default branch. time is in UTC.
     # * is a special character in YAML so you have to quote this string.
     # run every night at 10:00PM UTC.
     - cron:  '0 22 * * *'
+  push:
+    branches:
```
emiltin, line 18: We don't want to run tests on equipment on merges to branches. Instead we run when the tag 'testhub' is set on a PR, and on schedule.

emiltin:
Test are now running (since I added the 'testhub' tag to the PR), but for some reason the controller does not connect, or a process is blocking the address/port:

```
Aborting: Site did not connect within 30s


Finished in 30.06 seconds (files took 0.87623 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples

1 sentinel warnings:

  1 Errno::EADDRINUSE
```

emiltin:
> How can it be verified here that the controller actually goes through its start-up sequence if other states are accepted?

Is this related to a change in this PR or a more general question? I don't see changes in this PR related to the startup sequence.

pksoft72:
Aborting: Site did not connect within 30s ---> I was running another test on the same controller.

Startup sequence
- there are lot of commits int this PR, but most of them are taken back and only configuration of Cross RS4S was changed.
- I have solved startup sequence on Cross RS4S as logical state - so even pedestrian group without yellow bulb can report flashing yellow
https://github.com/rsmp-nordic/rsmp_validator/issues/465

.github/workflows/cross_rs4s.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: [ self-hosted, Linux, X64, cross-rs4s ]
     strategy:
       matrix:
-        core: ['3.1.2', '3.1.3', '3.1.4', '3.1.5', '3.2.0', '3.2.1', '3.2.2']
+        core: ['3.2.2', '3.1.2', '3.1.3', '3.1.4', '3.1.5', '3.2.0', '3.2.1'] # test last version first
```
emiltin, line 22: Good idea to start with the latest, but let's order them all as newest to oldest

config/cross_rs4s.yaml
```diff
@@ -34,8 +34,8 @@ components:
     CZ+76000=101DL002:
     CZ+76000=101DL003:
 items:
-  plans: [1,2,3,4,5]
-  traffic_situations: []
```
emiltin, line 38: The test "Traffic Situation is set with M0003" is skipped because no traffic situations configured here. Can some be configured, or does the controller not support this?

emiltin:
Is the controller an RS4S or RS4?

config/cross_rs4s.yaml
pksoft72: Controller accepts any traffic situation, behavior is dependent on traffic solution. 
I understand traffic situation as some information, which is used to select correct plan.
So it is possible in script(=user traffic logic) use current traffic situation to select different plan.

pksoft72:
> Is the controller an RS4S or RS4?

RSMP works on RS4S and RS4T. RS4 is old, not in production. RS4T is upgraded RS4.
RS4S and RS4T have identical firmware so their RSMP compatibility is the same.

emiltin:
Could you please squash to either a single commit, or a few relevant commits?

emiltin:
Core 3.2.2 and 3.2.0 succeeds, but all other versions fail, most with these tests failing:
```
rspec ./spec/site/tlc/modes_spec.rb:165 # Site::Traffic Light Controller Operational yellow flash can be activated with M0001
rspec ./spec/site/tlc/modes_spec.rb:180 # Site::Traffic Light Controller Operational yellow flash affects all signal groups
rspec ./spec/site/tlc/modes_spec.rb:228 # Site::Traffic Light Controller Operational dark mode can be activated with M0001
rspec ./spec/site/tlc/modes_spec.rb:243 # Site::Traffic Light Controller Operational yellow flash be used with a timeout of one minute
rspec ./spec/site/tlc/signal_groups_spec.rb:88 # Site::Traffic Light Controller Signal Group follow startup sequence after yellow flash
rspec ./spec/site/tlc/traffic_situations_spec.rb:11 # Site::Traffic Light Controller Traffic Situation is read with S0015
```

Do you know what's causing some versions to fails?

pksoft72:
> Could you please squash to either a single commit, or a few relevant commits?

I have changed it to single commit.

config/cross_rs4s.yaml
emiltin: No, traffic situations are something you configure on the device and can switch between, like signals plans. @otterdahl can you confirm this?

emiltin:
@pksoft72 Should we merge this now, and then open new issues for any remaining failing tests?

emiltin:
Replaced by #504 (same content, but rebased on main).

PR #502, opened 2025-01-21
update workflows to test newest core versions first

emiltin:


emiltin:
The Cross workflow file was not modified, as there's another PR with changes to the config for that device.

emiltin:
Not sure the ordering is respected by the runner, see https://github.com/actions/runner/issues/483.

PR #511, opened 2025-01-29
use ruby 3.4, update rsmp and other gems

emiltin:


emiltin:
Let's keep an eye on runners, since they need to update Ruby and gems.

PR #523, opened 2025-01-31
use sOc=true in status subscriptions

emiltin:


emiltin:
sOc was introduces in core 3.1.5. Older core version has no sOc flag, but if you set uRt=0, it means send when there's a change.

https://rsmp-nordic.github.io/rsmp_core/3.1.4/applicability/basic_structure.html#status-messages

So for core<3.1.5, we should set uRt=0, for core>=3.1.5, we should set uRt=0, sOc=true.

emiltin:
The Cartesio failures are due to no space left on device, so unrelated to this PR change.

PR #528, opened 2025-02-04
check that S0207 occupancy values are within -1..100

emiltin:


emiltin:
@anderspalm the ITC-3 is failing the updated S0207 tests.

emiltin:
@anderspalm note that this PR was merged and the ITC-3 now fails the involved tests, as detailed above.

Issue #531, opened 2025-02-18
Cartesio and TMacs runners cannot install gems


emiltin:
```
Warning: Cache not found for keys: setup-ruby-bundler-cache-v6-ubuntu-20.04-x64-ruby-3.4.2-wd-/opt/actions-runner/runner_cartesio/_work/rsmp_validator/rsmp_validator-with--without--only--Gemfile.lock-5b03d9b6237e67c8688a9bf1c2fdd3ae7d4542159ef8286461166ad4cc088754, setup-ruby-bundler-cache-v6-ubuntu-20.04-x64-ruby-3.4.2-wd-/opt/actions-runner/runner_cartesio/_work/rsmp_validator/rsmp_validator-with--without--only--Gemfile.lock-
/opt/hostedtoolcache/Ruby/3.4.2/x64/bin/bundle install --jobs 4
Fetching gem metadata from https://rubygems.org/.......
Fetching rake 13.2.1
Installing rake 13.2.1
...
Fetching thor 1.3.2
Installing thor 1.3.2
Error: The process '/opt/hostedtoolcache/Ruby/3.4.2/x64/bin/bundle' failed with exit code null
```

Perhaps a corrupted bundle cache? We might try this fix:
https://github.com/ruby/setup-ruby?tab=readme-ov-file#dealing-with-a-corrupted-cache

emiltin:
Some problem with TMacs

mariusrazvanplugaru:
Hi @emiltin . It is not clear, what operations should we do? and why can this happen?
BR,

emiltin:
See the PR #534. By setting `cache-version: 1` (increasing if needed) you can clear the bundler cache. I suspect this cache might have been corrupted during some update of gems, ruby or OS libraries. Let's see if this solves the problem.

emiltin:
It does not seem to work. bundler still exist with error code null (which is odd).

Does the TMac and Cartesio use the same runner machine?

mariusrazvanplugaru:
The last time i made something else, now i want to set the cache-version. Where or in which file i can set that ? 
BR

mariusrazvanplugaru:
I tried only, for now to execute only this commands:
gem cleanup
rm -rf /usr/local/lib/ruby/gems/3.3.0/cache

It could be enough?

Please you can restart the process ? 
BR,

mariusrazvanplugaru:
> It does not seem to work. bundler still exist with error code null (which is odd).
> 
[...]

PR #534, opened 2025-02-19
clear cartesio bundler cache

emiltin:
Fix for #531, the cartesio runner fails to install gems with bundler.

This PR clears the cache according to https://github.com/ruby/setup-ruby?tab=readme-ov-file#dealing-with-a-corrupted-cache.

### DIFF
--- a/.github/workflows/semaforica_cartesio.yaml
+++ b/.github/workflows/semaforica_cartesio.yaml
@@ -42,6 +42,7 @@ jobs:
     - name: Install Ruby and gems
       uses: ruby/setup-ruby@v1
       with:
+        self-hosted: true
         bundler-cache: true
         cache-version: 2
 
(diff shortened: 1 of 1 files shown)


emiltin:
@anezsergi after setting the self-hosted: true flag, the action reports:
```
Run ruby/setup-ruby@v1
Using 3.4 as input from file .tool-versions
Error: The current runner (ubuntu-20.04-x64) was detected as self-hosted because the self-hosted input was set.
In such a case, you should install Ruby in the $RUNNER_TOOL_CACHE yourself, for example using https://github.com/rbenv/ruby-build
You can take inspiration from this workflow for more details: https://github.com/ruby/ruby-builder/blob/master/.github/workflows/build.yml
$ ruby-build 3.4 /opt/actions-runner/runner_cartesio/_work/_tool/Ruby/3.4/x64
Once that completes successfully, mark it as complete with:
$ touch /opt/actions-runner/runner_cartesio/_work/_tool/Ruby/3.4/x64.complete
It is your responsibility to ensure installing Ruby like that is not done in parallel.
````

Let's first see if things work now. If now, I think we should install Ruby outselves, instead of using the prebuild Rubies that ruby/setup-ruby downloads. I would use a version manager to build ruby. e.g. mise.

PR #536, opened 2025-02-25
uses status subscription instead of request in S0207 test

emiltin:


emiltin:
The ITC-3 still returns S0207 occupancies that are far outside the range -1..100:
```
"s": "422,529,575,18846861,18846861,18846861,18521456,18846861,18846861,18846861"
```

https://github.com/rsmp-nordic/rsmp_validator/actions/runs/13515902028/job/37764425683#step:6:11695

PR #588, opened 2025-09-16
Test ITC-3 with SXL 1.2.1, instead of 1.0.15

emiltin:


emiltin:
@SwarcoPalm has the ITC-3 been reconfigured to use SXL 1.2.1?

emiltin:
ping @SwarcoPalm

PR #589, opened 2025-09-16
test satellite on all core versions

emiltin:


emiltin:
The controller send core versions [3.2.1, 3.2.0] in it's initial Version message, so connecting using other versions is not possible.
If other core version are supported, please add these to the core version array.

emiltin:
> Other Core versions have been added to the Version message.

nice, i'm rerunning the tests now

emiltin:
There are test faiilures on several core versions: https://github.com/rsmp-nordic/rsmp_validator/actions/runs/17761598644/job/50776895395?pr=589

PR #614, opened 2025-10-28
use rubocop branch of rsmp gem

emiltin:
A major code cleanup is going on in the rsmp gem. This PR pulls the a branch of the rsmp gem to verify the the tests workflows with external equipment.


emiltin:
The branch in the rsmp gem specifies the use of Ruby 3.4, but the Cross runner machine has 3.3, so can not run tests.
The Swarco runner machine does not have git, so cannot install a gem from a git repo.

But gem, Kapsch, Semaforica and TECSEN tests pass, showing that the rsmp gem and the validator itself works.

PR #640, opened 2026-04-10
Sus, package as gem, remove YARD

emiltin:
- [x] Switch from RSpec to Sus, which is written by the same author as the Async gem which we use extensively for concurrency, and fits better with what we need.
- [x] Package the validator as a Ruby gem with versioning, making it easier to install and update it.
- [x] A new rsmp-validator binary for running tests, which is a wrapper around sus with some config handling.
- [x] The rsmp gem now includes all the schema files, and the code to validate messages
- [x] Replaces complicated and annoying YARD doc generation with a custom solution, which is simpler and faster.
- [x] Adds a TLC Proxy with high-level abstraction for interacting with a remote TLC
- [x] dynamically pick core defintions to use in the sxl scheme validation
- [x] add workflow for testing the docker build


emiltin:
Good progress:
- Kapsch ETX tests pass for all cores version except 3.2.1.
- The gem TLC and supervisor passes all test.
- No rubocop violations
- New set of internal sus test all pass.

Next is investigating why other controllers and TMACS still fail tests.

